PR #5, opened 2020-12-04
created reader for finances and writer for calculation

Artificial-Friend:
IDE recommends me to change return value of writeTo to void by I think it is VERY useful to be able to get result done or not even though method already resulting by throwing exception or not.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,46 @@
 package core.basesyntax;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+
 public class WorkWithFile {
+    public static final String SUPPLY = "supply";
+    public static final String BUY = "buy";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        List<String> data = readFrom(fromFileName);
+        int supply = 0;
+        int buy = 0;
+        for (String line : data) {
+            if (line.contains(SUPPLY)) {
+                supply += Integer.parseInt(line.split(",")[1]);
+            } else if (line.contains(BUY)) {
+                buy += Integer.parseInt(line.split(",")[1]);
+            }
+        }
+        StringBuilder result = new StringBuilder().append("supply,").append(supply)
+                .append(System.lineSeparator()).append("buy,").append(buy)
+                .append(System.lineSeparator()).append("result,").append(supply - buy);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,46 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        List<String> data = readFrom(fromFileName);
+        int supply = 0;
+        int buy = 0;
+        for (String line : data) {
+            if (line.contains(SUPPLY)) {
+                supply += Integer.parseInt(line.split(",")[1]);
+            } else if (line.contains(BUY)) {
+                buy += Integer.parseInt(line.split(",")[1]);
+            }
```
dmytro-hryhoruk, line 21: if there is no third option, I leave just else without additional condition

dmytro-hryhoruk (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,46 @@
...
+
+    private List<String> readFrom(String fromFileName) {
+        try {
+            return Files.readAllLines(new File(fromFileName).toPath());
+        } catch (IOException e) {
+            throw new RuntimeException("ERROR: unable to find: " + fromFileName, e);
+        }
+    }
 
+    private boolean writeTo(String toFileName, StringBuilder message) {
```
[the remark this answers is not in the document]
Artificial-Friend: Agreed, my mistake.

src/main/java/core/basesyntax/WorkWithFile.java
Artificial-Friend: I've added it in case there will be more than two words. Refactored.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,45 @@
...
 public class WorkWithFile {
+    public static final String SUPPLY = "supply";
+    public static final String BUY = "buy";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        List<String> data = readFrom(fromFileName);
+        int supply = 0;
+        int buy = 0;
+        for (String line : data) {
+            if (line.contains(SUPPLY)) {
```
dmytro-hryhoruk, line 17:
```suggestion
            if (line.equals(SUPPLY)) {
```

dmytro-hryhoruk (approved):

PR #38, opened 2020-12-06
added read file, work with file, write file implementation

oroboro-cpu:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,56 @@
...
+            if (data[i].equals("supply")) {
+                supply += Integer.parseInt(data[i + 1]);
+            } else if (data[i].equals("buy")) {
+                buy += Integer.parseInt(data[i + 1]);
+            }
+        }
+        return stringBuilder.append("supply").append(",").append(supply)
+                .append(System.lineSeparator()).append("buy").append(",").append(buy)
+                .append(System.lineSeparator()).append("result").append(",")
+                .append(supply - buy).toString();
```
KseniiaMakarova, line 45: let's move this to the private method as well?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,56 @@
...
+
+    private String calculateResult(String value) {
+        int supply = 0;
+        int buy = 0;
+        StringBuilder stringBuilder = new StringBuilder();
+        String[] data = value.split(",");
+        for (int i = 0; i < data.length; i++) {
+            if (data[i].equals("supply")) {
+                supply += Integer.parseInt(data[i + 1]);
+            } else if (data[i].equals("buy")) {
```
KseniiaMakarova, line 38: let’s move values `,`, `supply`, `buy` to constant fields

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
 public class WorkWithFile {
+    private static final char COMMA = ',';
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        writeFile(toFileName,
+                writeReport(calculateSupply(readFile(fromFileName)),
+                        calculateBuy(readFile(fromFileName))));
```
KseniiaMakarova, line 19: should we call `readFile()` twice or can we save the result of the method call into a local variable?

KseniiaMakarova (changes requested):

PR #49, opened 2021-03-31
Homework

Nikber5:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,49 @@
...
+            throw new RuntimeException("Can't write data to file", e);
+        }
+    }
 
+    private String createReport(String fromFileName) {
+        File file = new File(fromFileName);
+        int supply = 0;
+        int buy = 0;
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
```
KseniiaMakarova, line 31: see checklist please - Remember, if you are using classes that implement an AutoCloseable interface, we should use it with try-with-resources.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,49 @@
...
+            String value = reader.readLine();
+            while (value != null) {
+                String[] valueContent = value.split(",");
+                if (valueContent[0].equals("supply")) {
+                    supply += Integer.parseInt(valueContent[1]);
+                } else if (valueContent[0].equals("buy")) {
+                    buy += Integer.parseInt(valueContent[1]);
+                }
```
KseniiaMakarova, line 39: see checklist - If you have strange strings or numbers in the code, it’s better to declare them as constants. The name of the constant should display this object’s purpose.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,49 @@
...
+        int supply = 0;
+        int buy = 0;
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
+            String value = reader.readLine();
+            while (value != null) {
+                String[] valueContent = value.split(",");
+                if (valueContent[0].equals("supply")) {
+                    supply += Integer.parseInt(valueContent[1]);
+                } else if (valueContent[0].equals("buy")) {
```
KseniiaMakarova, line 37: do we need `else if` or `else` will be enough?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
Nikber5: done

src/main/java/core/basesyntax/WorkWithFile.java
Nikber5: done

src/main/java/core/basesyntax/WorkWithFile.java
Nikber5: I left if else, because if we will change code in future there may be file with another operation type.

KseniiaMakarova: if in future there will be file with another operation type - we will need to refactor the code rather than write new `else if`

let's remove it for now

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,53 @@
...
+                    buy += Integer.parseInt(valueContent[AMMOUNT]);
+                }
+                value = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
+        }
+        return "supply," + supply + System.lineSeparator()
+                + "buy," + buy + System.lineSeparator()
+                + "result," + (supply - buy);
```
KseniiaMakarova, line 51: we can use our constants here

KseniiaMakarova (approved):
I also left one comment above

src/main/java/core/basesyntax/WorkWithFile.java
Nikber5: ok

src/main/java/core/basesyntax/WorkWithFile.java
Nikber5: Done

PR #60, opened 2021-04-02
Implemented readFromCsvFile method

Den4eg007:
Also added constant fields for usage in readFromCsvFile method

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,39 @@
 package core.basesyntax;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final int RESULT_ARRAY_LENGTH = 3;
+    private static final String[] CSV_CATEGORIES = new String[]{"buy", "supply"};
 
     public void getStatistic(String fromFileName, String toFileName) {
+    }
 
+    private String[] readFromCsvFile(String fromFileName) {
+        File fileFrom = new File(fromFileName);
+        List<String> csv_values;
+        int buyTotal = 0;
+        int supplyTotal = 0;
+
+        try {
+            csv_values = Files.readAllLines(fileFrom.toPath());
+            for (String currentLine : csv_values) {
+                if (currentLine.contains(CSV_CATEGORIES[0])) {
+                    buyTotal += Integer.parseInt(currentLine.split(",")[1]);
+                    continue;
+                }
+                supplyTotal += Integer.parseInt(currentLine.split(",")[1]);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,52 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        writeToCsvFile(toFileName, readFromCsvFile(fromFileName));
+    }
+
+    private String[] readFromCsvFile(String fromFileName) {
+        File fileFrom = new File(fromFileName);
+        List<String> csvValues;
+        int buyTotal = 0;
+        int supplyTotal = 0;
+        try {
```
RomanDubovskyi, line 23: checklist mistake (resources)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,52 @@
...
+            csvValues = Files.readAllLines(fileFrom.toPath());
+            for (String currentLine : csvValues) {
+                if (currentLine.contains(CSV_CATEGORIES[0])) {
+                    buyTotal += Integer.parseInt(currentLine.split(",")[1]);
```
RomanDubovskyi, line 27: "," and 1 are the magic numbers / strings, declare them as constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,52 @@
...
+                    continue;
+                }
+                supplyTotal += Integer.parseInt(currentLine.split(",")[1]);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
+        }
+        String[] calculatedTotal = new String[RESULT_ARRAY_LENGTH];
```
RomanDubovskyi, line 35: we don't need arr here, just use string and you won't need to run for each loop inside of writeToCsvFile method

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,50 @@
...
+    }
+
+    private String readFromCsvFile(String fromFileName) {
+        File fileFrom = new File(fromFileName);
+        List<String> csvValues;
+        int buyTotal = 0;
+        int supplyTotal = 0;
+        try {
+            csvValues = Files.readAllLines(fileFrom.toPath());
+            for (String currentLine : csvValues) {
```
RomanDubovskyi, line 25: More stable will be to call split before if once in case in future you will have categories buy,  buyPack, etc. But for now it's enough

RomanDubovskyi (approved):

PR #82, opened 2021-04-05
added solution

Veronika-Alekseienko:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,48 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    public static final String COMMA = ",";
```
RomanDubovskyi, line 10:
```suggestion
    public static final String CSV_SEPARATOR = ",";
```
let's name our variables according to their purpose and not value, they hold

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,48 @@
...
+    public static final String SUPPLY = "supply";
```
RomanDubovskyi, line 11: let's make these public, we don't use them outside the class, this way we're gonna follow  encapsulation principle

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,48 @@
...
 
     public void getStatistic(String fromFileName, String toFileName) {
+        int supply = 0;
+        int buy = 0;
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            for (String value = bufferedReader.readLine();
+                    value != null; value = bufferedReader.readLine()) {
+                String[] array = value.split(COMMA);
+                if (value.contains(SUPPLY)) {
+                    supply += Integer.parseInt(array[1]);
```
RomanDubovskyi, line 23: - let's declare 1 as AMOUNT_INDEX, to make code more informative

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,48 @@
...
+    public static final String RESULT = "result";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        int supply = 0;
+        int buy = 0;
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            for (String value = bufferedReader.readLine();
+                    value != null; value = bufferedReader.readLine()) {
+                String[] array = value.split(COMMA);
+                if (value.contains(SUPPLY)) {
```
RomanDubovskyi, line 22: as well, it's better to use value[OPERATION_INDEX].equals(SUPPLY)) it's gonna be easier to work with this code if we have more operations with similar name (this is not obligatory)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,48 @@
...
+                }
+            }
+            StringBuilder stringBuilder = new StringBuilder()
+                    .append(SUPPLY).append(COMMA).append(supply)
+                    .append(System.lineSeparator())
+                    .append(BUY).append(COMMA).append(buy)
+                    .append(System.lineSeparator())
+                    .append(RESULT).append(COMMA).append(supply - buy)
+                    .append(System.lineSeparator());
+            writeToFile(toFileName, stringBuilder.toString());
```
RomanDubovskyi, line 35: no you open connection inside of connection. Nested try blocks are considered to be a bad practice. Its' recommended to keep connection opened as short as possible.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,48 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fromFileName, e);
+        }
+    }
 
+    public void writeToFile(String toFileName, String report) {
```
RomanDubovskyi, line 41: if we don't use this method outside the class let's make it private

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
RomanDubovskyi: this comment is not fixed

RomanDubovskyi (changes requested):

RomanDubovskyi (approved):

PR #83, opened 2021-04-05
First solution

Oleksandr-Shulha:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+    public static final String BUY = "buy";
+    public static final String RESULT = "result";
+    public static final String SEPARATOR = ",";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        readDataFromFile(fromFileName);
+        createReport(toFileName, readDataFromFile(fromFileName));
+    }
+
+    public String readDataFromFile(String fileName) {
```
Sofasmile, line 21: should this method be public?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+        int supply = 0;
+        int buy = 0;
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName));
```
Sofasmile, line 25: let's use try with resources

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+            String stringLine = bufferedReader.readLine();
+            while (stringLine != null) {
+                if (stringLine.split(SEPARATOR)[0].equals(SUPPLY)) {
+                    supply += Integer.parseInt(stringLine.split(SEPARATOR)[1]);
+                } else {
+                    buy += Integer.parseInt(stringLine.split(SEPARATOR)[1]);
+                }
+                stringLine = bufferedReader.readLine();
+            }
+            bufferedReader.close();
```
Sofasmile, line 36: use close here is a bad practice because if we will have some exception our resource won't close

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+                    .append(BUY).append(SEPARATOR).append(buy)
+                    .append(System.lineSeparator())
+                    .append(RESULT).append(SEPARATOR).append(supply - buy)
+                    .append(System.lineSeparator()).toString();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from file " + fileName, e);
+        }
+    }
+
+    private void createReport(String fileName, String report) {
```
Sofasmile, line 48:
```suggestion
    private void writeReport(String fileName, String report) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+        try {
+            File fileReport = new File(fileName);
+            fileReport.createNewFile();
```
Sofasmile, line 51: let's remove first try block and createNewFile() operation

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+            try {
+                BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(fileName));
+                bufferedWriter.write(report);
+                bufferedWriter.close();
```
Sofasmile, line 55: the same about close

Sofasmile (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -14,15 +14,14 @@
     public static final String SEPARATOR = ",";
 
     public void getStatistic(String fromFileName, String toFileName) {
-        readDataFromFile(fromFileName);
-        createReport(toFileName, readDataFromFile(fromFileName));
+        createNewFile(toFileName);
```
Sofasmile, line 17: we don't need to create a new file, because a new file will be created when we call `new BufferedWriter(new FileWriter(fileName))`

Sofasmile (changes requested):
good job)

src/main/java/core/basesyntax/WorkWithFile.java
Oleksandr-Shulha: Its a qoute not from last commit?

Sofasmile: Я мала на увазі в попередньому коментарі "let's remove first try block and createNewFile() operation"
що нам потрібно видалити перший блок try і заодно з цим видалити і createNewFile(), адже він нам не потрібен

Oleksandr-Shulha: Done

Sofasmile (approved):

PR #90, opened 2021-04-05
add solution

iedenysenko:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,68 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String OPERATION_SUPPLY = "supply";
+    private static final String OPERATION_BUY = "buy";
+    private static final String OPERATION_RESULT = "result";
+    private static final String SPLIT_REGEX = ",";
+    private static final int INDEX_FOR_OPERATION_TYPE = 0;
+    private static final int INDEX_FOR_AMMOUNT = 1;
```
RomanDubovskyi, line 15:
```suggestion
    private static final int AMOUNT_INDEX = 1;
```
usually it's the common naming (of course it's not critical for code, just a heads up🙂)

RomanDubovskyi (approved):
Nice👍

src/main/java/core/basesyntax/WorkWithFile.java
iedenysenko: thanks

PR #92, opened 2021-04-05
First try

Nadhob:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,37 @@
...
 
+public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        int totalSupplies = 0;
+        int totalBuys = 0;
+        String line;
 
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            while ((line = bufferedReader.readLine()) != null) {
+                String[] singleRecord = line.split(",");
```
RomanDubovskyi, line 17: let's make "," 0 and 1 constants to improve readability

RomanDubovskyi (changes requested):
declare constants, rest LGTM

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,40 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    public static final int WORD_PART = 0;
+    public static final int NUMBER_PART = 1;
```
RomanDubovskyi, line 11:
```suggestion
    private static final int NUMBER_PART = 1;
    private static final String CSV_SEPARATOR = ",";
```
if we don't use constants outside let's make them private

RomanDubovskyi (approved):
left comment about constant declaration

PR #103, opened 2021-04-06
added solution

haniiev:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,60 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    public static final String SUPPLY = "supply";
+    public static final String BUY = "buy";
+    public static final String RESULT = "result";
+    public static final String COMMA = ",";
+    public static final int ZERO_POSITION = 0;
+    public static final int FIRST_POSITION = 1;
```
RomanDubovskyi, line 16:
```suggestion
    public static final int OPERATION_INDEX = 0;
    public static final int AMOUNT_INDEX = 1;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,60 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    public static final String SUPPLY = "supply";
+    public static final String BUY = "buy";
```
RomanDubovskyi, line 12: we don't use them outside the class, let's declare them as private

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,60 @@
...
+                    line != null;
+                    line = bufferedReader.readLine()) {
+                if (line.split(COMMA)[ZERO_POSITION].equals(SUPPLY)) {
+                    supply += Integer.parseInt(line.split(COMMA)[FIRST_POSITION]);
+                }
+                if (line.split(COMMA)[ZERO_POSITION].equals(BUY)) {
+                    buy += Integer.parseInt(line.split(COMMA)[FIRST_POSITION]);
+                }
+            }
+            writeToFile(toFileName, supply, buy);
```
RomanDubovskyi, line 33: you open connection ot a file inside another connection to file (try block inside try block). It's considered to be a bad practice. We need to close current resources before opening new one

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,60 @@
...
+
+        } catch (IOException e1) {
```
RomanDubovskyi, line 35: why e1?

RomanDubovskyi (changes requested):
alltogether it's good solution let's fix remaining mistakes

src/main/java/core/basesyntax/WorkWithFile.java
haniiev: but i use try with resources

RomanDubovskyi: yes, and at what point will resources be closed?

RomanDubovskyi (changes requested):
fix nested resources openning, rest LGTM

RomanDubovskyi (approved):

PR #110, opened 2021-06-05
Solution is done:

Dhakmar:
Implemented method getStatistic(String fromFileName, String toFileName) which causes method createAndGetReportFromFile and method writeReport.

Implemented method createAndGetReportFromFile(String fileName) which returns structured String report from market after working day.
Implemented method writeReport(String fileName, String report) which writes reports to file

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,46 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int INDEX_OF_VALUE = 1;
+    private static final String CVS_SEPARATOR = ",";
 
-    public void getStatistic(String fromFileName, String toFileName) {
+public void getStatistic(String fromFileName, String toFileName) {
+        writeReport(toFileName, createAndGetReportFromFile(fromFileName));
+    }
 
+    private String createAndGetReportFromFile(String fileName) {
+        int supply = 0;
+        int buy = 0;
+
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,46 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        writeReport(toFileName, createAndGetReportFromFile(fromFileName));
+    }
+
+    private String createAndGetReportFromFile(String fileName) {
+        int supply = 0;
+        int buy = 0;
+
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String readable;
```
KseniiaMakarova, line 22:
```suggestion
            String line;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,46 @@
...
+            while ((readable = bufferedReader.readLine()) != null) {
+                if (readable.contains("supply")) {
+                    supply += Integer.parseInt(readable.split(CVS_SEPARATOR)[INDEX_OF_VALUE]);
+                }
+                if (readable.contains("buy")) {
```
KseniiaMakarova, line 27: can we just use `else`?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,46 @@
...
+                    buy += Integer.parseInt(readable.split(CVS_SEPARATOR)[INDEX_OF_VALUE]);
```
KseniiaMakarova, line 28: let's save `readable.split(CVS_SEPARATOR)[INDEX_OF_VALUE]` into a variable and then reuse it

KseniiaMakarova (changes requested):
overall good job, added minor comments

KseniiaMakarova (approved):

PR #112, opened 2021-06-06
hw-5 added solution

OlesiaSmitiukh:
Implement method getStatistic using methods createReport(reading data from file .csv) 
and writeToFile (writing data to report file)

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,52 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int FIRST_INDEX = 0;
+    private static final int SECOND_INDEX = 1;
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,52 @@
...
+        int supplyAmount = 0;
+        int buyAmount = 0;
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = bufferedReader.readLine()) != null) {
+                String[] linesRead = line.split(CSV_SEPARATOR);
+                if (linesRead[FIRST_INDEX].equals(SUPPLY)) {
+                    supplyAmount += Integer.parseInt(linesRead[SECOND_INDEX]);
+                }
+                if (linesRead[FIRST_INDEX].equals(BUY)) {
```
KseniiaMakarova, line 31: can we just use `else`?

KseniiaMakarova (approved):
nice!

src/main/java/core/basesyntax/WorkWithFile.java
OlesiaSmitiukh: Yes, it will  be  working(I've just tried ). In this task we need sort two operation type. 
I think , what  I need to add a specific operation, it is better to use block "if". In block "else" could get other operations.

PR #123, opened 2021-06-07
Created several methods and implemented its logic

deniskichan:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,58 @@
...
+            throw new RuntimeException("Can't read from file", e);
+        }
+    }
 
+    private String createReport(String infoFromFile) {
+        int supplyAmount = 0;
+        int buyAmount = 0;
+        String[] splittedFile = infoFromFile.replace("\r", "").split("\n");
+        for (String splittedData : splittedFile) {
+            String[] splittedLine = splittedData.split(",");
```
KseniiaMakarova, line 45: it will be slightly better to save `,` to a constant as well

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,58 @@
...
+    private static final String RESULT = "result,";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        String infoFromFile = readFromFile(fromFileName);
+        String report = createReport(infoFromFile);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(report);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to file", e);
+        }
```
KseniiaMakarova, line 23: let's move this logic to a separate private method `writeToFile(String fileName, String content)`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,58 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int OPERATION_TYPE_INDEX = 0;
+    private static final int AMOUNT_INDEX = 1;
+    private static final String SUPPLY = "supply,";
+    private static final String BUY = "buy,";
+    private static final String RESULT = "result,";
```
KseniiaMakarova, line 14:
```suggestion
    private static final String SUPPLY = "supply";
    private static final String BUY = "buy";
    private static final String RESULT = "result";
    private static final String DELIMITER = ",";
```

KseniiaMakarova (changes requested):
very good job! left minor comments

KseniiaMakarova (approved):

PR #124, opened 2021-06-07
Program

testYurka:
Create program, which read data and write result. Logic was built in two methods.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -7,7 +7,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 
 public class WorkWithFile {
-    private static final int INDEX_OF_NUMERIC = 1;
+    private static final int INDEX_NUMERIC = 1;
 
     public void getStatistic(String fromFileName, String toFileName) {
         try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
@@ -26,11 +26,11 @@ public class WorkWithFile {
             while ((reader = bufferedReader.readLine()) != null) {
                 if (reader.contains("supply")) {
                     String [] inheritArray = reader.split(",");
-                    supply += Integer.parseInt(inheritArray[INDEX_OF_NUMERIC]);
+                    supply += Integer.parseInt(inheritArray[INDEX_NUMERIC]);
                 }
                 if (reader.contains("buy")) {
                     String [] inheritArray = reader.split(",");
-                    buy += Integer.parseInt(inheritArray[INDEX_OF_NUMERIC]);
+                    buy += Integer.parseInt(inheritArray[INDEX_NUMERIC]);
                 }
             }
         } catch (IOException e) {
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,43 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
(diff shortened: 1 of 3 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,43 @@
...
+            throw new RuntimeException("Can't write data to the file ", e);
+        }
+    }
+
+    private String readFromFile(String fileName) {
+        int supply = 0;
+        int buy = 0;
 
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String reader;
```
KseniiaMakarova, line 25:
```suggestion
            String line;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,43 @@
...
+            while ((reader = bufferedReader.readLine()) != null) {
+                if (reader.contains("supply")) {
+                    String [] inheritArray = reader.split(",");
+                    supply += Integer.parseInt(inheritArray[INDEX_NUMERIC]);
+                }
+                if (reader.contains("buy")) {
```
KseniiaMakarova, line 31: can we use `else`?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,43 @@
...
+                    String [] inheritArray = reader.split(",");
```
```suggestion
                    String[] splitArray = reader.split(",");
```
also, this logic is written twice

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,43 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int INDEX_NUMERIC = 1;
```
KseniiaMakarova, line 10:
```suggestion
    private static final int INDEX_OF_PRODUCT_COUNT = 1; 
```
let's try to use a more informative name

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,43 @@
...
+
+    private String readFromFile(String fileName) {
+        int supply = 0;
+        int buy = 0;
 
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String reader;
+            while ((reader = bufferedReader.readLine()) != null) {
+                if (reader.contains("supply")) {
+                    String [] inheritArray = reader.split(",");
```
KseniiaMakarova, line 28: it is better to save `,` to constant as well

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
testYurka: Yes, but as first I was not understand logic of the program))

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,43 @@
...
+    }
+
+    private String readFromFile(String fileName) {
+        int supply = 0;
+        int buy = 0;
 
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = bufferedReader.readLine()) != null) {
+                String [] splitArray = line.split(WORD_SEPARATOR);
```
KseniiaMakarova, line 29:
```suggestion
                String[] splitArray = line.split(WORD_SEPARATOR);
```

KseniiaMakarova (approved):

PR #153, opened 2021-08-19
Hw 4 solution

7777vova7777:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,49 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private final String sup = "supply";
```
RomanDubovskyi, line 11: let's make more informative names instead of  sup / buy / res

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,49 @@
...
+    private final String buy = "buy";
+    private final String res = "result";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        int addSupply = 0;
```
RomanDubovskyi, line 16:
```suggestion
        int supplySum = 0;
        int buySum
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,49 @@
...
+        int addBay = 0;
+        int result = 0;
+
+        File fileIn = new File(fromFileName);
+        File fileOut = new File(toFileName);
+
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileIn))) {
```
RomanDubovskyi, line 23: In general it's recommended to keep connection open the shortest possible time. So if we have any heavy computation inside of try with resources block it's better to move them outside and perform only read operations while connection is opened. After info is read we can then perform remaining operations. 
That's recommendation for future, for current task it's all good 🙂

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,49 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private final String sup = "supply";
+    private final String buy = "buy";
+    private final String res = "result";
 
     public void getStatistic(String fromFileName, String toFileName) {
```
RomanDubovskyi, line 15: let's split this method into smaller ones to increase readability

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
7777vova7777: Ok!

RomanDubovskyi (approved):

PR #187, opened 2021-08-24
created method which form report based on information from file, meth…

VitaliiMYKH:
…od which write this report into another file, and method which invoke previous two.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,63 @@
 package core.basesyntax;
 
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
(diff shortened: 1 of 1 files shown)


Resci (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+        generateReport(fromFileName);
+        writeReport(toFileName);
+    }
+
+    public void generateReport(String fromFileName) {
+        int supply = 0;
+        int buy = 0;
+        int result = 0;
+        File fromFile = new File(fromFileName);
+        List<String> fromFileInfo;
```
Resci, line 25: Move reading from file to the separate method

Resci (changes requested):

PR #208, opened 2021-09-18
Work with file solution

Angela1608:


src/test/java/core/basesyntax/WorkWithStatisticTest.java
```diff
@@ -7,7 +7,7 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 
-public class WorkWithFileTest {
+public class WorkWithStatisticTest {
```
RomanDubovskyi, line 10: we don't need to change anything in test classes

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,59 @@
...
+                int parseInt = Integer.parseInt(split[1]);
+                if (readAllLine.contains("supply")) {
+                    supply += parseInt;
+                } else {
+                    buy += parseInt;
+                }
+                result = supply - buy;
+            }
+            String s = String.valueOf(supply);
+            String b = String.valueOf(buy);
```
RomanDubovskyi, line 27: we don't need this conversion, append will work when you simply pass the value 
`.append(supply)`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,59 @@
...
-public class WorkWithFile {
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
 
+public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        try {
```
RomanDubovskyi, line 11: it's recommended to have try block as little as possible. It should only contain code that generates exception. Let's move forming of report outside of the try block

RomanDubovskyi (changes requested):

src/main/java/core/basesyntax/WorkWithStatistic.java
```diff
@@ -0,0 +1,56 @@
...
+import java.nio.file.Files;
+
+public class WorkWithStatistic {
+    public void getStatistic(String fromFileName, String toFileName) {
+        StringBuilder stringBuilder = new StringBuilder();
+        int supply = 0;
+        int buy = 0;
+        int result = 0;
+        try {
+            for (String readAllLine : Files.readAllLines(new File(fromFileName).toPath())) {
```
RomanDubovskyi, line 16: Краще зробити `try { String[] content = Files.readAllLines() } catch...` а потім проводити маніпуляції з контентом

RomanDubovskyi (approved):

PR #209, opened 2021-09-20
Wrote an assignment jv-work-with-file

ASelivonchyk:
implement method getStatistic according to the requirements

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,54 @@
 package core.basesyntax;
 
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,54 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
 
+public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        List<String> dayData = readFile(fromFileName);
+        int supplyAmount = countProduct(dayData, "supply");
```
RomanDubovskyi, line 12: ideally we would use single loop to initialize data to optimize performance

RomanDubovskyi (approved):
nice job with the structure

PR #211, opened 2021-09-23
created two methods getReport and writeReportToFile for implementatio…

mykolaskrypka:
…n of task

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,61 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,61 @@
...
+                    default:
+                        continue;
+                }
+                value = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
+        }
+
+        StringBuilder stringBuilder = new StringBuilder();
```
RomanDubovskyi, line 45: outside of the loop we can use simple String, StringBuilder is a must have only inside of the loop

RomanDubovskyi (approved):

PR #215, opened 2021-10-10
Implemented method 'getStatistic (String fromFileName, String toFileN…

SerhiiPet:
…ame)', which return data from the market after the working day.

There are the following parts:
- read from file: fromFileName;
- data calculations;
- write data to file: toFileName;

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,49 @@
 package core.basesyntax;
 
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,63 @@
...
+        try {
+            listFromFileName = Files.readAllLines(file.toPath());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
+        }
+        String readString = listFromFileName.toString();
+        return readString.substring(1, readString.length() - 1);
+    }
+
+    private int sumator(String strings, String article) {
```
RomanDubovskyi, line 36:
```suggestion
    private int countOperationTotal(String strings, String article) {
```

PR #222, opened 2021-10-18
1.Read file and done operation with numbers

Viktor-Stadnik:
2.Created 2 methods for print report and write file.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -9,43 +9,44 @@ import java.io.IOException;
 public class WorkWithFile {
 
     public void getStatistic(String fromFileName, String toFileName) {
-        int supplyCount = 0;
-        int buyCount = 0;
+        String data = readeFile(fromFileName);
+        String reportSend = report(data);
+        writeToFile(toFileName, reportSend);
+    }
 
-        //String report = reportPrint(supplyCount, buyCount);
-        // writeToFile(toFileName, report);
-
-        try {
-            BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
+    private String readeFile(String fromFileName) {
+        StringBuilder builder = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
             String row = reader.readLine();
             while (row != null) {
-                String name = row.substring(0, row.indexOf(","));
-                String number = row.substring(row.indexOf(",") + 1);
-                if (name.equals("supply")) {
-                    supplyCount += Integer.valueOf(number);
-                }
-                if (name.equals("buy")) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
 
     public void getStatistic(String fromFileName, String toFileName) {
+        int supplyCount = 0;
+        int buyCount = 0;
+
+        //String report = reportPrint(supplyCount, buyCount);
```
RomanDubovskyi, line 15: try not to push commented code

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+        // writeToFile(toFileName, report);
+
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
```
RomanDubovskyi, line 19: we need to put buffered reader into try with resources block as well, to close connection to file after we're done

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+            String row = reader.readLine();
+            while (row != null) {
+                String name = row.substring(0, row.indexOf(","));
```
RomanDubovskyi, line 22: you can use here split() operation as well

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+    }
+
+    public String reportPrint(int supply, int buy) {
+        StringBuilder report = new StringBuilder();
+        return (report.append("supply,").append(supply).append(System.lineSeparator())
+                .append("buy,").append(buy).append(System.lineSeparator())
+                .append("result,").append(supply - buy)).toString();
+    }
+
+    public void writeToFile(String toFileName, String reportPrint) {
```
RomanDubovskyi, line 47: if we don't use methods outside of the class let's make them private to follow encapsulation principle

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+            throw new RuntimeException("Can't read file" + fromFileName, e);
+        }
+
+        String report = reportPrint(supplyCount, buyCount);
+        writeToFile(toFileName, report);
+    }
+
+    public String reportPrint(int supply, int buy) {
+        StringBuilder report = new StringBuilder();
+        return (report.append("supply,").append(supply).append(System.lineSeparator())
```
RomanDubovskyi, line 42: we can use string concat outside of the loop. String builder is required only inside of the loop

PR #302, opened 2022-01-21
1st implementation of jv-work-with-file task

OleksiyTitov:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,75 @@
...
+        String[] data = readFromFile(fromFileName);
+        if (data != null) {
+            int countSupply = 0;
+            int countBuy = 0;
+            for (String dataLine : data) {
+                int indexOfComa = dataLine.indexOf(',') + 1;
+                if (dataLine.contains("supply")) {
+                    countSupply += Integer.parseInt(dataLine.substring(indexOfComa));
+                }
+                if (dataLine.contains("buy")) {
```
Max-Pochepets, line 23: strings should be constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,75 @@
...
 
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] data = readFromFile(fromFileName);
+        if (data != null) {
+            int countSupply = 0;
+            int countBuy = 0;
+            for (String dataLine : data) {
+                int indexOfComa = dataLine.indexOf(',') + 1;
+                if (dataLine.contains("supply")) {
+                    countSupply += Integer.parseInt(dataLine.substring(indexOfComa));
```
Max-Pochepets, line 21: lets not use substring, but use split instead

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,75 @@
...
+                }
+                if (dataLine.contains("buy")) {
+                    countBuy += Integer.parseInt(dataLine.substring(indexOfComa));
+                }
+            }
+            String result = "supply,"
```
Max-Pochepets, line 27: we should use StringBuilder for string concatenation

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,75 @@
...
+                    + countSupply
+                    + '/'
+                    + "buy,"
+                    + countBuy
+                    + '/'
```
Max-Pochepets, line 32: let's separate with `System.lineSeparator()` and pass to `writeToFile` a ready to use string, not an array

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,75 @@
...
+            }
+        }
     }
+
+    private String[] readFromFile(String fileName) {
+        File file = new File(fileName);
+        try (BufferedReader reader = new BufferedReader(new FileReader(file))) {
+            StringBuilder stringBuilder = new StringBuilder();
+            String line = reader.readLine();
+            if (line != null) {
```
Max-Pochepets, line 59: for this task lets assume that `line` will never be null. we don't need this check

Max-Pochepets (changes requested):

Max-Pochepets (approved):

Max-Pochepets (changes requested):
fix build issues

Max-Pochepets (approved):

PR #304, opened 2022-01-22
work-with-file

TonyHut:
1)created method readFile which reads file and return info in the array.
2)created method makeReport which makes report and returns him like a String.
3)method fileWrite takes the String and writs into the new file.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,56 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,56 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+
 public class WorkWithFile {
+    static final String SEPARATOR = ",";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        writeFile(makeReport(readFile(fromFileName)), toFileName);
+
```
Resci, line 15:
```suggestion
```

Resci (approved):

PR #312, opened 2022-01-23
solved hw and make some methods

AprioriKS:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,76 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private int supply = 0;
+    private int buy = 0;
+    private int result = 0;
+
+    public int getSupply() {
```
Max-Pochepets, line 15: there is no need in getters and setters

Max-Pochepets (changes requested):
read checklist once more

Max-Pochepets (approved):

PR #317, opened 2022-01-24
added solution

Navijava:


Max-Pochepets (changes requested):
read checklist once more: logic division, constants, try-with-resources. and mind the naming, everything should be named in full, not in shortcuts e.g.
str - Bad
string - Good

PR #341, opened 2022-02-09
created two variables type File for reading and writing. In construct…

kate-chern:
…ion try-catch created BufferedReader and read all information from a file using readLine and save to List<String>. After that BufferedReader closed. Created array String and in the outer loop created int array for calculating and build stringBuilder where we save result calculation. In other construction try-catch created BufferedWriter and wrote result to file.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,55 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
 public class WorkWithFile {
+    private final List<String> allLines = new ArrayList<>();
 
     public void getStatistic(String fromFileName, String toFileName) {
-
+        File file1 = new File(fromFileName);
+        File file2 = new File(toFileName);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+                }
+                stringBuilder.append(reportString[j]).append(",").append(reportInt[j])
+                .append(System.lineSeparator());
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data to file", e);
+        }
+        BufferedWriter bufferedWriter;
+        try {
+            bufferedWriter = new BufferedWriter(new FileWriter(file2));
```
Resci, line 46: Чому ти не використовуєш `try-with-resources`?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
 public class WorkWithFile {
+    private final List<String> allLines = new ArrayList<>();
 
     public void getStatistic(String fromFileName, String toFileName) {
-
+        File file1 = new File(fromFileName);
+        File file2 = new File(toFileName);
```
Resci, line 17:
```suggestion
        File inputFile = new File(fromFileName);
        File outputFile = new File(toFileName);
```

Resci (changes requested):

PR #350, opened 2022-03-10
Dev

DenysShl:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,61 @@
...
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
 public class WorkWithFile {
 
     public void getStatistic(String fromFileName, String toFileName) {
+        Map<String, Integer> hashMap = new LinkedHashMap<>();
```
KseniiaMakarova, line 16: can we use HashMap here?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,61 @@
...
+        File file = new File(fromFileName);
+        List<String> stringList = null;
+        try {
+            stringList = Files.readAllLines(file.toPath());
+        } catch (IOException e) {
+            throw new RuntimeException("File does not exist: " + fromFileName, e);
+        }
+        for (String s : stringList) {
```
KseniiaMakarova, line 24: please use appropriate naming

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,61 @@
...
+            String[] arr = s.split(",");
```
KseniiaMakarova, line 25: use constants please

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,61 @@
...
+            if (hashMap.containsKey(arr[0])) {
+                hashMap.put(arr[0], hashMap.get(arr[0]) + Integer.valueOf(arr[1]));
+            } else {
+                hashMap.put(arr[0], Integer.valueOf(arr[1]));
+            }
+        }
+        hashMap.put("result", hashMap.get("supply") - hashMap.get("buy"));
```
KseniiaMakarova, line 32: same here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,61 @@
...
+        saveToFile(toFileName, getFormater(hashMap));
+    }
+
+    private String getFormater(Map<String, Integer> hashMap) {
+        StringBuilder stringBuilder = new StringBuilder();
+        hashMap.entrySet().stream()
+                .sorted(Map.Entry.comparingByValue(Comparator.reverseOrder()))
```
KseniiaMakarova, line 39: why are we sorting like this? will it produce correct report in all cases?

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
DenysShl: Make without using Hash Map?

KseniiaMakarova: I meant, you could use HashMap instead of LinkedHashMap

KseniiaMakarova (approved):

PR #378, opened 2022-03-29
Methods for data transition were created

oleksandr-protsenko:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,38 @@
...
+                }
+                value = reader.readLine();
+            }
+            writeStatistic(writer, buy, supply);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file ", e);
+        }
+    }
 
+    public void writeStatistic(BufferedWriter writer, int buy, int supply) throws IOException {
```
```suggestion
    private void writeStatistic(int buy, int supply) {
```
let's open writer separately and catch IOException in this method

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,38 @@
...
+            String value = reader.readLine();
+            while (value != null) {
+                if (value.split(",")[0].equals(SUPPLY)) {
+                    supply += Integer.parseInt(value.split(",")[1]);
+                } else {
+                    buy += Integer.parseInt(value.split(",")[1]);
+                }
+                value = reader.readLine();
+            }
+            writeStatistic(writer, buy, supply);
```
KseniiaMakarova, line 26: move this line (where you write) out of the try block

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,41 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fromFileName, e);
+        }
+        writeStatistic(toFileName, buy, supply);
+    }
 
+    private void writeStatistic(String toFileName, int buy, int supply) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName))) {
+            String str = "supply," + supply + System.lineSeparator() + "buy," + buy
```
KseniiaMakarova, line 33: `str` is not a good name

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,41 @@
...
+                    + System.lineSeparator() + "result,"
+                    + (supply - buy);
```
KseniiaMakarova, line 35: please don't use plain concatenation

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #393, opened 2022-04-06
add solution jv-work-with-file

SerhiiPankov:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,36 @@
 package core.basesyntax;
 
-public class WorkWithFile {
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
 
+public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
MaliukDaria, line 9: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file#divide-program-logic): Divide program logic

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,36 @@
...
+            int result = supply - buy;
+            String[] outputString = new String[] {"supply," + supply + System.lineSeparator(),
+                    "buy," + buy + System.lineSeparator(),
+                    "result," + result};
+            toFile.createNewFile();
+            for (String string: outputString) {
+                Files.writeString(toFile.toPath(),string, StandardOpenOption.APPEND);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can'n work with file");
```
MaliukDaria, line 33: not infornative exception message, it is not clear where was the problem in read from file or in write to file. And you should add file name to exception message

MaliukDaria (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
MaliukDaria: Lets move reading from file and write to file to separate methods

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,46 @@
...
+        String[] strings;
+        try {
+            strings = Files.readAllLines(fromFile.toPath()).toArray(new String[0]);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fromFileName, e);
+        }
+        try {
+            toFile.createNewFile();
+            WorkWithFile workWithFile = new WorkWithFile();
+            String[] outputString = workWithFile.outputString(strings);
```
MaliukDaria, line 21:
```suggestion
            String[] outputString = outputString(strings);
```

we could call methods of this class just by their name, we don't need to create istance of this class

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #394, opened 2022-04-06
Added and implemented necessary methods

vasiliy-sheretko:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+            String line = bufferedReader.readLine();
+            String[] split;
+            while (line != null) {
+                split = line.split(",");
+                switch (split[OPERATION_TYPE_INDEX]) {
+                    case "buy":
+                        buy += Integer.parseInt(split[AMMOUNT_INDEX]);
+                        break;
+                    default:
+                        supply += Integer.parseInt(split[AMMOUNT_INDEX]);
```
MaliukDaria, line 34:
```suggestion
                    case "buy":
                        buy += Integer.parseInt(split[AMMOUNT_INDEX]);
                        break;
                    case "supply":
                        supply += Integer.parseInt(split[AMMOUNT_INDEX]);
                        break;
```

default it for handle not expected values, but we expect to see "supply"

MaliukDaria (changes requested):

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #398, opened 2022-04-11
Sort task

AlexOriobel:
WorkWithFile.java: Class which has 2 methods readFile and writeFile;

CalculationStatistics.java: Class which calculate statistics this class was created to separate file Reader-Writer logic and calculation logic to simplify interaction and read of cod.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,41 @@
 package core.basesyntax;
 
+import core.basesyntax.calculation.CalculationStatistics;
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+
 public class WorkWithFile {
+    private ArrayList<String> str = new ArrayList<>();
+    private CalculationStatistics calculation = new CalculationStatistics();
 
     public void getStatistic(String fromFileName, String toFileName) {
+        readFile(fromFileName);
+        writeFile(toFileName);
+    }
 
+    private void readFile(String fileNameRead) {
+        String text;
+        try (FileReader fileFrom = new FileReader(fileNameRead)) {
+            BufferedReader bufferedReader = new BufferedReader(fileFrom);
+            while ((text = bufferedReader.readLine()) != null) {
+                str.add(text);
+            }
+            str.trimToSize();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fileNameRead, e);
+        }
+    }
+
+    private void writeFile(String fileNameWrite) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(fileNameWrite))) {
+            bufferedWriter.write(calculation.getNumber(str));
+            bufferedWriter.flush();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to the file " + fileNameWrite, e);
+        }
     }
 }
--- a/src/main/java/core/basesyntax/calculation/CalculationStatistics.java
+++ b/src/main/java/core/basesyntax/calculation/CalculationStatistics.java
@@ -1,0 +1,25 @@
+package core.basesyntax.calculation;
+
+import java.util.ArrayList;
+
+public class CalculationStatistics {
+    private static final String BUY_OPERATION = "buy";
+    private static final String SUPPLY_OPERATION = "supply";
+
+    public static String getNumber(ArrayList<String> str) {
+        int supplyOperation = 0;
+        int buyOperation = 0;
+        for (String strO : str) {
+            if (strO.trim().startsWith(SUPPLY_OPERATION)) {
+                supplyOperation += Integer.parseInt(strO.substring(strO.indexOf(',') + 1));
+            } else if (strO.trim().startsWith(BUY_OPERATION)) {
+                buyOperation += Integer.parseInt(strO.substring(strO.indexOf(',') + 1));
+            } else {
+                break;
+            }
+        }
+        return "supply," + supplyOperation + System.lineSeparator()
+                + "buy," + buyOperation + System.lineSeparator()
+                + "result," + (supplyOperation - buyOperation);
+    }
+}


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,41 @@
...
 public class WorkWithFile {
+    private ArrayList<String> str = new ArrayList<>();
+    private CalculationStatistics calculation = new CalculationStatistics();
 
     public void getStatistic(String fromFileName, String toFileName) {
+        readFile(fromFileName);
+        writeFile(toFileName);
+    }
+
+    private void readFile(String fileNameRead) {
```
Nick97-git, line 20: let's return List<String> here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,41 @@
...
+            while ((text = bufferedReader.readLine()) != null) {
+                str.add(text);
+            }
+            str.trimToSize();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fileNameRead, e);
+        }
+    }
 
+    private void writeFile(String fileNameWrite) {
```
Nick97-git, line 33: pass as the second parameter formed report

src/main/java/core/basesyntax/calculation/CalculationStatistics.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax.calculation;
+
+import java.util.ArrayList;
+
+public class CalculationStatistics {
```
Nick97-git, line 5: remove this class nad move its method in WorkWithFile class

src/main/java/core/basesyntax/calculation/CalculationStatistics.java
```diff
@@ -0,0 +1,25 @@
...
+    private static final String BUY_OPERATION = "buy";
```
Nick97-git, line 6: it's redundant remove these constants

src/main/java/core/basesyntax/calculation/CalculationStatistics.java
```diff
@@ -0,0 +1,25 @@
...
+        int supplyOperation = 0;
+        int buyOperation = 0;
+        for (String strO : str) {
+            if (strO.trim().startsWith(SUPPLY_OPERATION)) {
+                supplyOperation += Integer.parseInt(strO.substring(strO.indexOf(',') + 1));
+            } else if (strO.trim().startsWith(BUY_OPERATION)) {
+                buyOperation += Integer.parseInt(strO.substring(strO.indexOf(',') + 1));
+            } else {
+                break;
+            }
```
Nick97-git, line 19: can we remove this else?

src/main/java/core/basesyntax/calculation/CalculationStatistics.java
```diff
@@ -0,0 +1,25 @@
...
+        }
+        return "supply," + supplyOperation + System.lineSeparator()
```
Nick97-git, line 21: use StringBuilder instead of String concatenation

src/main/java/core/basesyntax/calculation/CalculationStatistics.java
```diff
@@ -0,0 +1,25 @@
...
+public class CalculationStatistics {
+    private static final String BUY_OPERATION = "buy";
+    private static final String SUPPLY_OPERATION = "supply";
+
+    public static String getNumber(ArrayList<String> str) {
+        int supplyOperation = 0;
+        int buyOperation = 0;
+        for (String strO : str) {
+            if (strO.trim().startsWith(SUPPLY_OPERATION)) {
+                supplyOperation += Integer.parseInt(strO.substring(strO.indexOf(',') + 1));
```
Nick97-git, line 14: let's use split method instead of substring, it looks messy

src/main/java/core/basesyntax/calculation/CalculationStatistics.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class CalculationStatistics {
+    private static final String BUY_OPERATION = "buy";
+    private static final String SUPPLY_OPERATION = "supply";
+
+    public static String getNumber(ArrayList<String> str) {
+        int supplyOperation = 0;
+        int buyOperation = 0;
+        for (String strO : str) {
+            if (strO.trim().startsWith(SUPPLY_OPERATION)) {
```
Nick97-git, line 13: do you need to call trim()?

Nick97-git (changes requested):

src/main/java/core/basesyntax/calculation/CalculationStatistics.java
AlexOriobel: > use StringBuilder instead of String concatenation
Ther`s no loop, so what a reason use StringBuilder?

Nick97-git: because it makes your program faster

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,57 @@
...
+            while ((text = bufferedReader.readLine()) != null) {
+                str.add(text);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fileNameRead, e);
+        }
+        return str;
+    }
+
+    private void writeFile(String fileNameWrite, List<String> str) {
```
Nick97-git, line 31:
```suggestion
    private void writeFile(String fileName, List<String> lines) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,57 @@
...
+        for (String line : str) {
+            String[] splitFile = line.split(",");
+            if (splitFile[0].equals("buy")) {
+                buyOperation += Integer.parseInt(splitFile[1]);
+            }
+            if (splitFile[0].equals("supply")) {
+                supplyOperation += Integer.parseInt(splitFile[1]);
+            }
+        }
+        StringBuilder b = new StringBuilder();
```
Nick97-git, line 52:
```suggestion
        StringBuilder reportBuilder = new StringBuilder();
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,57 @@
...
+    private void writeFile(String fileNameWrite, List<String> str) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(fileNameWrite))) {
+            bufferedWriter.write(getNumber(str));
+            bufferedWriter.flush();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to the file " + fileNameWrite, e);
+        }
+    }
 
+    public String getNumber(List<String> str) {
```
Nick97-git, line 40:
```suggestion
    public String getReport(List<String> lines) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,57 @@
...
+        int supplyOperation = 0;
+        int buyOperation = 0;
+        for (String line : str) {
+            String[] splitFile = line.split(",");
```
Nick97-git, line 44:
```suggestion
            String[] splittedLine = line.split(",");
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,57 @@
...
+            if (splitFile[0].equals("buy")) {
+                buyOperation += Integer.parseInt(splitFile[1]);
+            }
+            if (splitFile[0].equals("supply")) {
+                supplyOperation += Integer.parseInt(splitFile[1]);
```
Nick97-git, line 49: create constants from each of indexes, for example 0 -> `private static final int TYPE_OF_OPERATION_INDEX = 0;`

Nick97-git (changes requested):

src/main/java/core/basesyntax/calculation/CalculationStatistics.java
AlexOriobel: According to oracle documents there's no difference in this case and cod is shorter if we useing +  it's easier to read 
https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.18.1

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,60 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final int CONDITION_FOR_INDEX = 0;
+    private static final int SPLIT_INDEX = 1;
```
Nick97-git, line 13:
```suggestion
    private static final int TYPE_OF_OPERATION_INDEX = 0;
    private static final int QUANTITY_INDEX = 1;
```

Nick97-git (approved):

PR #402, opened 2022-04-18
Completed

olesyakhalakk:


MaliukDaria (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,49 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private int supply = 0;
+    private int buy = 0;
```
MaliukDaria, line 11: I don't think we need this parameters as class level, lent make them local. You can return array with 2 elements from readFromFile method

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,49 @@
...
 
     public void getStatistic(String fromFileName, String toFileName) {
+        readFromFile(fromFileName);
+        StringBuilder reportBuilder;
+        reportBuilder = new StringBuilder();
+        reportBuilder.append("supply,").append(supply).append(System.lineSeparator())
+                .append("buy,").append(buy).append(System.lineSeparator())
+                .append("result,").append(supply - buy)
+                .append(System.lineSeparator());
+        String report = reportBuilder.toString().trim();
```
MaliukDaria, line 21: lets move logic with report creation to separate method

MaliukDaria (changes requested):

MaliukDaria (approved):
Good Job 👍🏻

PR #419, opened 2022-05-14
Implement methods: getStatistic, getDataFromFile, writeDataFromFile, …

Sm0kk:
…isInteger, getSumSupply, getSumBuy. Read data from file, parce, calculate and write to the file.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,81 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] dataFromFile = getDataFromFile(fromFileName);
+        int sumSupply = getSumOfSupply(dataFromFile);
+        int sumBuy = getSumOfBuy(dataFromFile);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,81 @@
...
+        }
+    }
+
+    private boolean isInteger(String data) {
+        try {
+            Integer.parseInt(data);
+            return true;
+        } catch (NumberFormatException e) {
+            return false;
+        }
```
KostyaNikiforov, line 59: Let's use next implementation:
```suggestion
        return data.matches('\\d+');
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,81 @@
...
+    private static final String BUY = "buy";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] dataFromFile = getDataFromFile(fromFileName);
+        int sumSupply = getSumOfSupply(dataFromFile);
+        int sumBuy = getSumOfBuy(dataFromFile);
+        StringBuilder report = new StringBuilder();
+        report.append(SUPPLY + "," + sumSupply).append(System.lineSeparator())
+                .append(BUY + "," + sumBuy).append(System.lineSeparator())
+                .append("result,").append(sumSupply - sumBuy);
```
KostyaNikiforov, line 22: Let's use only one type of concatenation

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
Sm0kk: That's really looks better) But regular expressions its yet darkness for me))

KostyaNikiforov (approved):
:fire:

PR #443, opened 2022-05-24
4 methods were implemented in WorkWithFile class

olyaklipa:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,66 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    public static final String SUPPLY = "supply";
```
liliia-ponomarenko, line 11:
```suggestion
    private static final String SUPPLY = "supply";
```
constants must have a private access modifier

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,66 @@
...
+    public static final String BUY = "buy";
+    public static final int ITEM = 0;
```
liliia-ponomarenko, line 13:
```suggestion
    public static final int ACTION = 0;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,66 @@
...
+                supplyCount = supplyCount + Integer.parseInt(splitRecord[VALUE]);
+            }
+            if (splitRecord[ITEM].equals(BUY)) {
+                buyCount = buyCount + Integer.parseInt(splitRecord[VALUE]);
+            }
+        }
+        int result = supplyCount - buyCount;
+        return "supply," + supplyCount + SEPARATOR
+                + "buy," + buyCount + SEPARATOR
+                + "result," + result;
```
liliia-ponomarenko, line 55: you can use constants to make the report

liliia-ponomarenko (changes requested):
good solution 🚀

liliia-ponomarenko (approved):
good job  👏

PR #481, opened 2022-07-14
Passed all the tests and checked by mvn clean package

VitaliiShkliarskyi:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,48 @@
...
 public class WorkWithFile {
+    private static final int INDEX_OF_NAME = 0;
+    private static final int INDEX_OF_VALUE = 1;
 
     public void getStatistic(String fromFileName, String toFileName) {
-
+        File fromFile = new File(fromFileName);
+        StringBuilder builder = new StringBuilder();
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fromFile));
```
MaliukDaria, line 18: I think we need to use try with resources here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,48 @@
...
+            String value = reader.readLine();
+            while (value != null) {
+                builder.append(value).append("; ");
+                value = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
```
MaliukDaria, line 25:
```suggestion
            throw new RuntimeException("Can't read file " + fromFileName, e);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,48 @@
...
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int INDEX_OF_NAME = 0;
+    private static final int INDEX_OF_VALUE = 1;
 
     public void getStatistic(String fromFileName, String toFileName) {
```
MaliukDaria, line 14: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file): Divide program logic

MaliukDaria (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task, and correct all the points described there. The mentor will not check the work until the checklist points are corrected

PR #506, opened 2022-07-29
Implement getStatistic method also implements such methods: writeInRe…

veers-droid:
…sult(which creates .csv file and writes in the result in correct form), createReport (return String formatted to mach the requirements for output result) and convertToInt

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,63 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+
 public class WorkWithFile {
+    private static final String OPERATION_BUY = "buy";
+    private static final String DEMILITER = ",";
+    private final StringBuilder report = new StringBuilder();
+    private int totalBuy = 0;
+    private int totalSupply = 0;
 
     public void getStatistic(String fromFileName, String toFileName) {
+        File file = new File(fromFileName);
 
+        try (FileReader fileReader = new FileReader(file);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,63 @@
...
+            }
+        } catch (Exception e) {
+            throw new RuntimeException("Cant get data from file - " + fromFileName);
+        }
+        writeInResult(toFileName);
+    }
+
+    private int convertToInt(String num) {
+        return Integer.parseInt(num);
+    }
```
KseniiaMakarova, line 39: do we need a separate method for this?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,63 @@
...
+        return report.toString();
+    }
+
+    private void writeInResult(String toFile) {
+        String reportToWrite = createReport();
+        File inputFile = new File(toFile);
 
+        try (FileWriter fileWriter = new FileWriter(inputFile)) {
+            fileWriter.write(reportToWrite);
+        } catch (Exception e) {
```
KseniiaMakarova, line 59: same - IOException is better?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,63 @@
...
+            throw new RuntimeException("Can't write data to the file" + toFile, e);
```
KseniiaMakarova, line 60:
```suggestion
            throw new RuntimeException("Can't write data to the file " + toFile, e);
```

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #509, opened 2022-07-30
Create method reading from csv and writing results to file

Alexhqolma:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,70 @@
...
+import java.util.Iterator;
+
 public class WorkWithFile {
 
     public void getStatistic(String fromFileName, String toFileName) {
+        File file = new File(fromFileName);
+        File outputFile = new File(toFileName);
+        StringBuilder outputResult = new StringBuilder();
+        try {
+            BufferedReader fileByLine = new BufferedReader(new FileReader(file));
```
dSkarzh, line 20: please go through common mistakes checklist and refactor your code

dSkarzh (changes requested):

dSkarzh (changes requested):
Please think how you can redo your task, because this logic that you have now is way too complex, redundant and completely not justified
Good solutions use only one loop, instead of list, sets and maps
Also your code is very hard to read and review
Don't forget to check yourself using common mistakes list

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,66 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.ArrayList;
+
 public class WorkWithFile {
```
dSkarzh, line 11:
```suggestion
public class WorkWithFile {
    private static final int OPERATION_INDEX = 0;
    private static final int AMOUNT_INDEX = 1;
    private static final String BUY_OPERATION = "buy";
    private static final String SUPPLY_OPERATION = "supply";

```

dSkarzh: constants for better readibility

dSkarzh (changes requested):
Nice start, needs to be refactored!

dSkarzh (approved):

PR #535, opened 2022-08-02
was implemented method getStatistic

valentyn-slutskyi:


src/main/java/core/basesyntax/Reader.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+
+public class Reader {
```
dSkarzh, line 8: better way to separate logic is to create several methods in `WorkWithFile`, because right now your `Reader` not only reads data, but also calculates statistics

src/main/java/core/basesyntax/Reader.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.FileReader;
+import java.io.IOException;
+
+public class Reader {
+    public int[] readFromFile(String fileName) {
+        int supply = 0;
+        int buy = 0;
+        File file = new File(fileName);
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(file));
```
dSkarzh, line 14: use try with resources
go through common mistakes list and check yourself once more

src/main/java/core/basesyntax/Writer.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class Writer {
+    public void writeToFile(int[] data, String fileName) {
+        File file = new File(fileName);
```
dSkarzh, line 10: same thing as in reader, about dividing of logic

dSkarzh (changes requested):

src/main/java/core/basesyntax/Writer.java
valentyn-slutskyi: Yes, there is the same line in different methods, but it received different arguments: input and output fileName. Changed names of arguments by more understandible

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,54 @@
...
+        File file = new File(fromFileName);
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
+            String row = bufferedReader.readLine();
+            while (row != null) {
+                String[] data = row.split(",");
+                if (data[0].equals("supply")) {
+                    supply += Integer.parseInt(data[1]);
+                } else {
+                    buy += Integer.parseInt(data[1]);
+                }
```
dSkarzh, line 36:
```suggestion
                if (data[OPERATION_INDEX].equals(OPERATION_SUPPLY)) {
                    supply += Integer.parseInt(data[AMOUNT_INDEX]);
                } else {
                    buy += Integer.parseInt(data[AMOUNT_INDEX]);
                }
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,54 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
```
dSkarzh, line 11:
```suggestion
public class WorkWithFile {
    private static final int OPERATION_INDEX = 0;
    private static final int AMOUNT_INDEX = 0;
    private static final String OPERATION_BUY = "buy";
    private static final String OPERATION_SUPPLY = "supply";

```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,54 @@
...
 
     public void getStatistic(String fromFileName, String toFileName) {
+        WorkWithFile workWithFile = new WorkWithFile();
+        int[] data = workWithFile.readFromFile(fromFileName);
```
dSkarzh, line 14:
```suggestion
        int[] data = readFromFile(fromFileName);
```
you can access method even without object creation, since we are in the same class

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,54 @@
...
+        StringBuffer stringBuffer = new StringBuffer();
+        stringBuffer.append("supply,").append(data[0])
+                    .append(System.lineSeparator())
+                    .append("buy,").append(data[1])
+                    .append(System.lineSeparator())
+                    .append("result,").append(data[0] - data[1]);
```
dSkarzh, line 20:
```suggestion
        stringBuffer.append(OPERATION_SUPPLY).append(data[0])
                    .append(System.lineSeparator())
                    .append(OPERATION_BUY).append(data[1])
                    .append(System.lineSeparator())
                    .append("result,").append(data[0] - data[1]);
```

dSkarzh (changes requested):

dSkarzh (approved):

PR #536, opened 2022-08-02
The suggested solution was added.

DioStesso1:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,43 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int FIRST_INDEX = 0;
```
dSkarzh, line 10:
```suggestion
    private static final int OPERATION_INDEX = 0;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,43 @@
...
+    private static final int AMOUNT_INDEX = 1;
+    private static final String LINE_SEPARATOR = ",";
```
dSkarzh, line 12:
```suggestion
    private static final String COMMA = ",";
```

dSkarzh (approved):
Great job!

PR #560, opened 2022-08-12
Do a homework

Vladislav-Dzyadevich:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,87 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 public class WorkWithFile {
+    private List<String> listOfFilesData = new ArrayList<>();
```
aaananas, line 13:
```suggestion
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,87 @@
...
 
     public void getStatistic(String fromFileName, String toFileName) {
+        StringBuilder dataToWrite = new StringBuilder();
+        readDataFromFile(fromFileName);
```
aaananas, line 17:
```suggestion
        List<String> dataFromFile = readDataFromFile(fromFileName);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,87 @@
...
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            while ((line = reader.readLine()) != null) {
+                listOfFilesData.addAll(Arrays.asList(line.split(",")));
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from file: " + fileName, e);
+        }
+    }
+
+    public String[] generateArrayWithCorrectData(List<String> data, String[] result) {
```
aaananas, line 55:
```suggestion
    public String createReport(List<String> data) {
```
The method should look like this. It should take data from a file, parse it, and create a string report, that later is going to be just put in the output file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,87 @@
...
+            int firstSum,
+            int secondSum) {
+        for (int i = 0; i < listOfFilesData.size(); i++) {
+            if (listOfFilesData.get(i).equals(finalArrayOfData[0])) {
+                firstSum += Integer.parseInt(listOfFilesData.get(i + 1));
+            }
+            if (listOfFilesData.get(i).equals(finalArrayOfData[1])) {
+                secondSum += Integer.parseInt(listOfFilesData.get(i + 1));
+            }
+        }
```
aaananas, line 83: basically this logic has to be in createReport() method

aaananas (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
Vladislav-Dzyadevich: Why bring this field inside a method if it is used in several methods?

src/main/java/core/basesyntax/WorkWithFile.java
Vladislav-Dzyadevich: To assign the value of this method to a field, you must change the return type of the readDataFromFile method, which makes no sense

src/main/java/core/basesyntax/WorkWithFile.java
Vladislav-Dzyadevich: As far as I understood clean code correctly, we have to divide the logic into small methods which are responsible for a single action, so I don't think that makes sense, also the length of a method should not exceed 15 lines

Vladislav-Dzyadevich:
The proposed options for changing the code are not clear to me, please describe the advantages of the proposed changes

src/main/java/core/basesyntax/WorkWithFile.java
aaananas: Because you divide logic between methods
Perfectly you only read those values in readFromFile, and then parse it in createReport
That is 2 times

From readDataFromFile you expect something to be returned. You should think about how to divide logic, and think about each part separately.
ReadFromFile - means read from file, nothing more or less

src/main/java/core/basesyntax/WorkWithFile.java
aaananas: it makes sense

src/main/java/core/basesyntax/WorkWithFile.java
aaananas: generateSumForEachElementOfCorrectDataArray
what is "sum for each element"
what is "correct data array"
why in methods name you have only array, but as parameters you pass list, some array "finalArray"?, firstSum, secondSum - This doesn't make sense

aaananas: Imagine looking at this code 2 months later. Would you still understand?

aaananas (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 public class WorkWithFile {
+    private StringBuilder dataToWrite = new StringBuilder();
 
     public void getStatistic(String fromFileName, String toFileName) {
+        List<String> dataFromFile = readDataFromFile(fromFileName);
+        String finalArrayOfData = createReport(dataFromFile, new String[2],0,0);
```
Nick97-git, line 17:
```suggestion
        List<String> lines = readDataFromFile(fromFileName);
        String statistic = createReport(dataFromFile, new String[2],0,0);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
+
```
Nick97-git, line 18: remove empty line

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
+        writeDataToFile(toFileName, finalArrayOfData);
+    }
+
+    public List<String> readDataFromFile(String fromFileName) {
+        List<String> res = new ArrayList<>();
```
Nick97-git, line 23:
```suggestion
        List<String> lines = new ArrayList<>();
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
+        String line;
```
Nick97-git, line 24: move this line into try

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
+            while ((line = reader.readLine()) != null) {
+                res.addAll(Arrays.asList(line.split(",")));
```
Nick97-git, line 27: why do you need to split here? add whole line

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(dataToWrite);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't find a file: " + fileName, e);
+        }
+    }
 
+    public String createReport(List<String> data, String[] result,
+                               int sumOfTheValuesOfTheFirstElements,
+                               int sumOfTheValuesOfTheSecondElements) {
```
Nick97-git, line 45:
```suggestion
    public String createReport(List<String> lines) {
```
 you don't need to pass all these parameters
create two variables for a total sum of supply and a total sum of buy
go through all lines using for loop and split each line and get all needed data to decide to which variable you need to add a number in a line

Nick97-git (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
Vladislav-Dzyadevich: To write with commas

src/main/java/core/basesyntax/WorkWithFile.java
Vladislav-Dzyadevich: I don't understand this solution in what aspects is my approach worse?

Nick97-git: if you have questions ask in chats, mentors don't answer here

Nick97-git (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
Vladislav-Dzyadevich: Updated

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,63 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 public class WorkWithFile {
+    private StringBuilder dataToWrite = new StringBuilder();
```
aaananas, line 13: remove this global variable
you should create StringBuilder in the place where you use it

aaananas (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
Vladislav-Dzyadevich: ok

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
 public class WorkWithFile {
```
aaananas, line 13:
```suggestion
```

aaananas (approved):

PR #567, opened 2022-08-25
Implemented method getStatistic

ValeriyM250:
Processed Errors which can be appeared during reading from file and processed Errors which can be appeared during creating and writing in file

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -9,7 +9,7 @@ import java.util.List;
 import java.util.Objects;
 
 public class WorkWithFile {
-    private static final int ARRAY_LENGTH = 2;
+    public static final int ARRAY_LENGTH = 2;
 
     public void getStatistic(String fromFileName, String toFileName) {
         writeToFile(createReport(readFromFile(fromFileName)), toFileName);
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,46 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,46 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    public static final int OPERATION_INDEX = 0;
+    public static final int COUNT_INDEX = 1;
 
     public void getStatistic(String fromFileName, String toFileName) {
-
+        File file = new File(fromFileName);
```
aaananas, line 16: image
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected” and stop homework review

aaananas (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,54 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    public static final int OPERATION_INDEX = 0;
+    public static final int COUNT_INDEX = 1;
 
     public void getStatistic(String fromFileName, String toFileName) {
```
aaananas, line 15: this method should have logic divided like this
```suggestion
    public void getStatistic(String fromFileName, String toFileName) {
        String dataFromFile = readFromFile(fromFileName);
        String report = createReport(dataFromFile);
        writeToFile(report, toFileName);
```

aaananas (changes requested):

zachar-chorny (changes requested):
All logic should be in createReport. And readFromFile should simply return the list of lines from the file, I recommend to use Files.readAllLines for this

kozhukhovsky (changes requested):
The previous remark has not been corrected. You need to divide the logic into methods. Anastasia described everything perfectly.

Rommelua (changes requested):
You should fix (or answer to) all mentor's comments before re-request new review. Be attentive and do not hesitate to contact us in Slack if you do not understand something.

image

PR #578, opened 2022-09-09
data from market

dima-senko:


Rommelua (changes requested):
Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected.
See [Common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file)
Divide program logic

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,57 @@
...
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                dateWithFile.append(value).append(System.lineSeparator());
+                value = bufferedReader.readLine();
+            }
+            example = getExample(dateWithFile.toString());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file" + fromFileName, e);
+        }
+        getWriterWithFile(example, toFileName);
```
aaananas, line 27: this method should look like this:
`List<String> dataFromFile = readFromFile(fromFileName);`
`String report = getReport(dataFromFile);`
`writeToFile(report, toFileName);`

method readFromFile() should only read data from file
getReport() should create report, parse everything? calculate... and return ready report as string
writeToFile() should oly get some string and write it to file
those methods should be private

aaananas (changes requested):

kozhukhovsky (approved):

PR #579, opened 2022-09-18
method getStatistic was implemented

kateryna-mykh:


aaananas (approved):

aaananas (commented):
For future, cretae a new branch and implement your solution there

PR #582, opened 2022-09-21
Implemented method getStatistic(String fromFileName, String toFileNam…

igorgalenko:
…e) which returns some data from market after working day

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,47 @@
 package core.basesyntax;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final String COMMA_SEPARATOR = ",";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        try {
+            writeDataToFile(calculateData(readDataFromFile(fromFileName)),toFileName);
+        } catch (IOException e) {
+            throw new RuntimeException("Error while working with files");
+        }
+    }
 
+    private List<String> readDataFromFile(String fileName) throws IOException {
+        return Files.readAllLines(Path.of(fileName));
+    }
+
+    private List<String> calculateData(List<String> listData) {
+        int totalBuy = 0;
+        int totalSupply = 0;
+
+        for (String line : listData) {
+            String[] strings = line.split(COMMA_SEPARATOR);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,47 @@
...
+                totalBuy += Integer.parseInt(strings[1]);
+            } else {
+                totalSupply += Integer.parseInt(strings[1]);
+            }
+        }
+
+        List<String> result = new ArrayList<>(3);
+        result.add("supply" + COMMA_SEPARATOR + totalSupply);
+        result.add("buy" + COMMA_SEPARATOR + totalBuy);
+        result.add("result" + COMMA_SEPARATOR + (totalSupply - totalBuy));
```
hronograf, line 40: [Use constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file#use-constants)
(for `supply`, `buy`, `result`)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,47 @@
...
+import java.util.List;
+
 public class WorkWithFile {
+    private static final String COMMA_SEPARATOR = ",";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        try {
+            writeDataToFile(calculateData(readDataFromFile(fromFileName)),toFileName);
+        } catch (IOException e) {
+            throw new RuntimeException("Error while working with files");
```
hronograf, line 16: in this case it will be difficult to find why error was thrown, because it isn't informative enough 
but we can do something like `throw new RuntimeException("...", e)`, so stack trace will be more informative

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,47 @@
...
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final String COMMA_SEPARATOR = ",";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        try {
+            writeDataToFile(calculateData(readDataFromFile(fromFileName)),toFileName);
```
hronograf, line 14: one-line solutions looks cool, but can be not so readable
but in this case it is 🆒

hronograf (changes requested):
Good job!
Let's improve a bit)

hronograf (approved):
LGTM! 🔥

PR #587, opened 2022-09-29
have feeling this is trash solution (((

alex-anofriev:


src/main/java/core/basesyntax/CountSupplyAndBuy.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class CountSupplyAndBuy {
+    private final String replaceChar = "";
+    private final int initialAmount = 0;
```
hronograf, line 5: in this case `replaceChar` and `initialAmount` may be redundant
we can see their meaning from the context

src/main/java/core/basesyntax/CountSupplyAndBuy.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class CountSupplyAndBuy {
```
hronograf, line 3: name of a class should be a noun
like `SupplyAndBuyCounter`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -2,7 +2,21 @@
 
 public class WorkWithFile {
```
hronograf, line 4: [Please don’t add redundant empty lines to your code.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#please-dont-add-redundant-empty-lines-to-your-code)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -2,7 +2,21 @@
...
 
-    public void getStatistic(String fromFileName, String toFileName) {
+    public static final String SUPPLY = "supply";
+    public static final String BUY = "buy";
+    public static final String COMMA = ",";
+    public static final String SPACE = " ";
+    private final BuffReader buffReader = new BuffReader();
+    private final CountSupplyAndBuy countSupplyAndBuy = new CountSupplyAndBuy();
+    private final CreateReport createReport = new CreateReport();
```
hronograf, line 11: good)

src/main/java/core/basesyntax/CreateReport.java
```diff
@@ -0,0 +1,11 @@
...
+
+public class CreateReport {
+
+    public String[] reportForFile(int allSupply, int allBuy, String res) {
+        return new String[] {WorkWithFile.SUPPLY + WorkWithFile.COMMA
+                + allSupply + System.lineSeparator(),
+                WorkWithFile.BUY + WorkWithFile.COMMA + allBuy + System.lineSeparator(),
+                res};
+    }
+}
```
hronograf, line 11: I think it would be better to move this logic to `WorkWithFile` class
after that we can make our constants `private`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -2,7 +2,21 @@
...
 
+    public void getStatistic(String fromFileName, String toFileName) {
+        String fileRead = buffReader.readFromFile(fromFileName);
+        String[] splittedWithSpace = fileRead.split(SPACE);
+        int totalSupply = countSupplyAndBuy.countSupply(splittedWithSpace);
+        int totalBuy = countSupplyAndBuy.countBuy(splittedWithSpace);
+        String result = "result" + COMMA + (totalSupply - totalBuy);
```
hronograf, line 18: we can introduce a constant for `"result"`

hronograf (changes requested):
Good!
interesting division of program logic)

src/main/java/core/basesyntax/BuffWriter.java
```diff
@@ -0,0 +1,46 @@
+package core.basesyntax;
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class BuffWriter {
```
hronograf, line 9: if it is `BuffWriter`, why it has `readFromFile` file method?)

src/main/java/core/basesyntax/BuffWriter.java
```diff
@@ -0,0 +1,46 @@
...
+    private String fileToRead;
+    private String fileToWrite;
+
+    public void writeInFile() {
+        CreateReportMessage message = new CreateReportMessage();
```
hronograf, line 14: it would be better to have arguments like `String fileToWrite` and `String text`
`writeInFile` should only write data to the file, not creating report and do other things

src/main/java/core/basesyntax/BuffWriter.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
+public class BuffWriter {
+    private String fileToRead;
+    private String fileToWrite;
```
hronograf, line 11: it would be more convenient if we could pass this fields as arguments to our methods

src/main/java/core/basesyntax/CountBuysAndSupply.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class CountBuysAndSupply {
```
hronograf, line 3: `Count` is a verb, but class names should be nouns

src/main/java/core/basesyntax/SplitAllSpaces.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class SplitAllSpaces {
+
+    public String[] splitSpaces() {
+        String a = WorkWithFile.buffWriter.readFromFile();
+        return a.split(WorkWithFile.SPACE);
+    }
+}
```
hronograf, line 9: let's remove this class, it is redundant

src/main/java/core/basesyntax/CountBuysAndSupply.java
```diff
@@ -0,0 +1,28 @@
...
+
```
hronograf, line 4: redundant line

src/main/java/core/basesyntax/CreateReportMessage.java
```diff
@@ -0,0 +1,20 @@
...
+    private final String totalBuys = WorkWithFile.BUY + WorkWithFile.COMMA
+            + allBuys + System.lineSeparator();
+    private final String reportMessage = WorkWithFile.REPORT + WorkWithFile.COMMA
+            + (allSupply - allBuys);
+    private final String[] reportArray = {totalSupply, totalBuys, reportMessage};
+
+    public String[] getReportArray() {
+        return reportArray;
+    }
+}
```
hronograf, line 20: let's remove `getReportArray` and introduce method that gets `String text` parameter
and returns a report, then we can use it in the `WorkWithFile` class

hronograf (changes requested):
Ok
and I think that previous solution was better

alex-anofriev:
My goal was to make every method in new class and try to unite their work

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -2,7 +2,23 @@
 
 public class WorkWithFile {
 
-    public void getStatistic(String fromFileName, String toFileName) {
+    public static final String SUPPLY = "supply";
+    public static final String BUY = "buy";
+    public static final String COMMA = ",";
+    public static final String SPACE = " ";
+    public static final int DEFAULT_NUMBER = 0;
```
hronograf, line 9: it would be better to move these constants to classes where they are used and make them `private`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -2,7 +2,23 @@
...
+    private final BufferReadWrite buffer = new BufferReadWrite();
+    private final BuyAndSupplyCounter buyAndSupplyCounter = new BuyAndSupplyCounter();
+    private final ReportMessageCreate reportMessageCreate = new ReportMessageCreate();
 
+    public void getStatistic(String fromFileName, String toFileName) {
+        int supplyAmount;
+        int buyAmount;
```
hronograf, line 16: why do we declare variables here and initialize them later?
can we move declaration to the initialization?

src/main/java/core/basesyntax/WorkWithFile.java
hronograf: check redundant lines in all classes

hronograf (changes requested):
now it is better

hronograf (changes requested):
it will be approved after you fix the build:

image

KostyaNikiforov (approved):
🔥

PR #589, opened 2022-09-30
read from and write to csv file

Bodddd:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,51 @@
...
+            throw new RuntimeException("Can't read data from " + fromFile.getName(), e);
+        }
+        return new int[]{supply,buy};
+    }
 
+    private void writeInfoToCsvFile(int supply, int buy, File toFile) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFile))) {
+            bufferedWriter.write("supply," + supply + System.lineSeparator());
+            bufferedWriter.write("buy," + buy + System.lineSeparator());
+            bufferedWriter.write("result," + (supply - buy) + System.lineSeparator());
```
liliia-ponomarenko, line 46: Maybe better pass the full report in this method and call the method write() one time?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,51 @@
...
+        } catch (IOException e) {
+            e.printStackTrace();
```
liliia-ponomarenko, line 48: ```e.printStackTrace()``` bad practice

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,51 @@
...
+        try {
+            List<String> lines = Files.readAllLines(fromFile.toPath());
+            for (String info : lines) {
+                String[] data = info.split(SEPARATED_VALUES);
+                if (data[TEXT_INFO_INDEX].equals("supply")) {
+                    supply += Integer.parseInt(data[1]);
+                } else {
+                    buy += Integer.parseInt(data[1]);
+                }
+            }
```
liliia-ponomarenko, line 35: Do we really need to keep all this code in a block try? Let’s move reading from the file to a separate private method.

liliia-ponomarenko (changes requested):
good solution🔥 let’s improve it 😉

ChumakAn (approved):

PR #600, opened 2022-10-03
created solution for workWithFile

Bohdan20b:


ChumakAn (changes requested):
Method should have only one purpose and be short and not complicated. Try to simplify this and make some smaller methods and call them in getStatistic().
Also, such things as `System.lineSeparator()` ,  "," and others can be constants, not magic numbers/strings

PR #603, opened 2022-10-03
Implemented method getStatistic(String fromFileName, String toFileNam…

PavloPolovyi:
…e) which return some data from market after working day.


ChumakAn (approved):
Well done! LGTM

PR #606, opened 2022-10-04
getStatistic implemented

OleksiyMasaltsev:


KostyaNikiforov (changes requested):
Your solution looks good for codewars, but it is kind of difficult to support in production.
The method `getStatistic` do too many things. It would be better to separate your solution into three private methods, which will read data from file, generate report and write data to file.

KostyaNikiforov (approved):
Good job):fire:

PR #609, opened 2022-10-04
Implemented getStatistic method

DeonOni:


ChumakAn (changes requested):
try to simplify the getStatistic method and make smaller methods with one purpose and use them in getStatistic(). It will help you make your code more readable. Also remember about constants. For example, System.lineSeparator(), "," , "supply" and others can be constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Can't get data", e);
+        }
+    }
+
+    public String getData(String fromFileName) {
+        File inputData = new File(fromFileName);
+        StringBuilder stringBuilderInput = new StringBuilder();
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(inputData));
```
ChumakAn, line 34: think of try-with-resources

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+            int value = reader.read();
+            while (value != -1) {
+                stringBuilderInput.append((char) value);
+                value = reader.read();
+            }
```
ChumakAn, line 39: any other ideas how to implement this?

ChumakAn (changes requested):
Good, let's try to make it better

DeonOni:
We need try-with-resources for close resources, but here we have nothing to close, or I do not understand something ?

ChumakAn (approved):
LGTM

PR #610, opened 2022-10-04
Implemented getStatistic() method in WorkWithFie class

NSay0:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,68 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
 public class WorkWithFile {
+    private static final String DELIMITER = ",";
+    private static final String LINE_SEPARATOR = System.lineSeparator();
```
KostyaNikiforov, line 12: This is redundant, you can just use the method directly

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,68 @@
...
+            csvContent = readCsvFile(Path.of(fromFileName));
+            report = generateReport(csvContent);
+        } catch (IOException | NumberFormatException e) {
+            throw new RuntimeException("Unable to read file: " + e);
+        }
+        try {
+            writeToCsv(Path.of(toFileName), report);
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to write file: " + e);
+        }
```
KostyaNikiforov, line 32: Let's move exception handling as close as possible to the place where it can be thrown

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,68 @@
...
+                csvContent.put(key, newValue);
+                continue;
+            }
+            csvContent.put(
+                    line.split(DELIMITER)[KEY_POSITION],
+                    Integer.parseInt(line.split(DELIMITER)[VALUE_POSITION])
+            );
+        }
+        return csvContent;
+    }
```
KostyaNikiforov, line 52: This method do to much things. The method name is `readCsvFile`, but i think it can not be possible to reuse it with any Csv

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
NSay0: System.lineSeparator() removed from constants;
+fixed in new commit

src/main/java/core/basesyntax/WorkWithFile.java
NSay0: try-catch bloks moved to read/write methods.
+fixed in new commit

src/main/java/core/basesyntax/WorkWithFile.java
NSay0: Sum logic moved to generateReport() method.
+fixed in new commit

KostyaNikiforov (approved):
🔥

PR #611, opened 2022-10-04
reading to and writing from file implemented, report needs review

TymurProkhorov:
Can`t realise how to separate different logical parts to different methods, so now I`ve just marked it by comments.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -9,39 +9,21 @@ import java.io.IOException;
 
 public class WorkWithFile {
     private static final String LINE_SEPARATOR = System.lineSeparator();
-    private static final String DATA_SEPARATOR = ", ";
+    private static final String DATA_SEPARATOR = "\\W+";
     private static final String SUPPLY_CONSTANT_FOR_RESULT = "supply,";
     private static final String BUY_CONSTANT_FOR_RESULT = "buy,";
     private static final String RESULT_CONSTANT = "result,";
-    private String dataFromFile;
-    private String report;
-
-    public String getDataFromFile() {
-        return dataFromFile;
-    }
-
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,59 @@
 package core.basesyntax;
 
-public class WorkWithFile {
+import java.io.*;
```
ChumakAn, line 3: common mistake 
image

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,59 @@
...
 
+public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        // read from file
+        File file = new File(fromFileName);
+        StringBuilder stringBuilder = new StringBuilder();
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
```
ChumakAn, line 11: what about try-with-resources?

ChumakAn (changes requested):
try to make separate methods with this code, you can divide it in methods with the similar parts as you've added comments 
Also remember about reading the common mistakes and avoiding them. Also you have some magic numbers and strings, make them contants

src/main/java/core/basesyntax/WorkWithFile.java
TymurProkhorov: I saw it, but couldnt fix. Thanks for help)

ChumakAn (changes requested):
Your code didn't pass the tests, you can see the detailed logs in the Java CI build failure report. When you call createReport method, you don't save the result of report to any variable and don't pass it to the writeToFile method, as a result, you have nothing to write into the file

ChumakAn (approved):
LGTM

PR #618, opened 2022-10-04
Jv work with file

vazonchik:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,71 @@
...
+        return new String[]{SUPPLY + COMMA + sumOfSupply,
+                BUY + COMMA + sumOfBy,
+                RESULT + COMMA + (sumOfSupply - sumOfBy)};
+    }
 
+    private String[] getArrFromFile(String fromFileName) throws IOException {
+        try (FileReader fl = new FileReader(fromFileName)) {
+            StringBuilder enteredData = new StringBuilder();
+            BufferedReader reader = new BufferedReader(fl);
+            String str;
```
ChumakAn, line 63: it's better not to use abbreviations like fl, str

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,71 @@
...
+                sumOfBy += Integer.parseInt(parsedS[PARSED_INDEX_1]);
+            }
+        }
+        return new String[]{SUPPLY + COMMA + sumOfSupply,
+                BUY + COMMA + sumOfBy,
+                RESULT + COMMA + (sumOfSupply - sumOfBy)};
+    }
 
+    private String[] getArrFromFile(String fromFileName) throws IOException {
+        try (FileReader fl = new FileReader(fromFileName)) {
```
ChumakAn, line 60: it's better to catch the exception in this method, not in the getStatic

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,71 @@
...
+        String[] arrWithResult = getArrWithResult(arrFromFile);
+        try {
+            putResultToFile(toFileName, arrWithResult);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write into the file " + toFileName);
+        }
+    }
+
+    private void putResultToFile(String toFileName, String[] arrWithResult) throws IOException {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
```
ChumakAn, line 36: same here, you are writing try block, but don't catch anything, it's better to catch exception here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,71 @@
...
+            arrFromFile = getArrFromFile(fromFileName);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file " + fromFileName);
+        }
+        String[] arrWithResult = getArrWithResult(arrFromFile);
+        try {
+            putResultToFile(toFileName, arrWithResult);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write into the file " + toFileName);
+        }
```
ChumakAn, line 32: ideally, it should look like 
(if you catch exceptions in the implementation of the method, instead of writing 'throws' in the method signature)
```suggestion
        arrFromFile = getArrFromFile(fromFileName);
        String[] arrWithResult = getArrWithResult(arrFromFile);
        putResultToFile(toFileName, arrWithResult);
```

ChumakAn (changes requested):
Good! Let's make it even better👍

ChumakAn (approved):
LGTM

PR #624, opened 2022-10-05
Hw work with file

Veronika-Tretiak:


ChumakAn (changes requested):
Well done! Looks good! Only minor comment - **System.lineSeparator()** can be a constant too

ChumakAn (approved):
LGTM

PR #625, opened 2022-10-05
Read all data from input csv file, I create a report and write it to …

Mar4uk21:
…newFile

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,55 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private int supply = 0;
+    private int buy = 0;
+    private int result = 0;
+    private final String split = ",";
+    private int index = 1;
 
     public void getStatistic(String fromFileName, String toFileName) {
-
+        File file = new File(fromFileName);
+        StringBuilder builder = new StringBuilder();
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
+            String value = reader.readLine();
+            while (value != null) {
+                builder.append(value).append(System.lineSeparator());
+                value = reader.readLine();
+            }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+            writer.write("supply," + supply + System.lineSeparator());
+            writer.flush();
+            writer.write("buy," + buy + System.lineSeparator());
+            writer.flush();
+            writer.write("result," + result + System.lineSeparator());
+            writer.flush();
+            writer.close();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file");
+        }
```
KostyaNikiforov, line 53: This method is too big and difficult to read and support. Let's separate your solution into three private method: for writing to the file, for reading, and for report generation

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private int supply = 0;
+    private int buy = 0;
+    private int result = 0;
+    private final String split = ",";
+    private int index = 1;
```
KostyaNikiforov, line 15: These fields are method implementation details. Seems like it won't work properly if you call the getStatistic method several times. Move them to the method level

KostyaNikiforov (changes requested):
Also, i recommend you use constants for all static values in your code. Like: 1, 2, ", ", etc

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,68 @@
...
+            String value = reader.readLine();
+            while (value != null) {
+                builder.append(value).append(System.lineSeparator());
+                value = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read this file", e);
+        }
+        return builder;
+    }
```
KostyaNikiforov, line 32: Let's pass the text you need to write to the params

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #628, opened 2022-10-05
HW work with file

Matolch:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,57 @@
...
+                builder.append((char) value);
+                value = reader.read();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Cant read file", e);
+        }
+        String sentence = builder.toString();
+        int sumSupply = 0;
+        int sumBuy = 0;
+        String[] splitSentense = sentence.split("\\W+");
```
ChumakAn, line 32: "\\W+" can be a constant

ChumakAn (changes requested):
Please divide the method into smaller methods and combine them in getStatistic. One method should have one purpose

src/main/java/core/basesyntax/WorkWithFile.java
ChumakAn: the same for the System.lineSeparator()
Please, be attentive and use constants

ChumakAn (approved):

PR #632, opened 2022-10-05
homwork tastk

Volodymyr-Mykychak:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,46 @@
...
+            }
+            int result = supplyNum - buyNum;
+            return "supply," + supplyNum + SEPARATOR + "buy," + buyNum
+                    + SEPARATOR + "result," + result;
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t read file" + e);
+        }
     }
+
 }
```
KostyaNikiforov, line 46: It is a good practice to separate different parts of a solution into private methods. 
It would be if you had:
- a method that reads data from file.
- a method that writes data to file.
- and a method that transforms input string into output (crate a statistic)

KostyaNikiforov (changes requested):

hronograf (changes requested):
common mistake
image

also, please, read Kostya's comment once again

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,83 @@
...
+        for (String line : report) {
+            try {
+                Files.write(path, line.getBytes(), StandardOpenOption.WRITE);
+            } catch (IOException e) {
+                throw new RuntimeException("Can't write in file", e);
+            }
+        }
+    }
+
+    public String[] getReport(Path path) {
```
hronograf, line 45:
```suggestion
    private String[] getReport(Path path) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,83 @@
...
+        saveReport(report, pathToFileName);
+    }
+
+    private void createFile(Path path) {
+        try {
+            Files.createFile(path);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't create file", e);
+        }
+    }
```
hronograf, line 32: do we need this method?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,83 @@
...
+import java.nio.file.StandardOpenOption;
+
 public class WorkWithFile {
+    private static final int OPERATION_TYPE_INDEX = 0;
+    private static final int AMOUNT_INDEX = 1;
+    private static final String SUPPLY_VALUE = "supply";
+    private static final String BUY_VALUE = "buy";
+    private static final String RESULT_VALUE = "result";
+    private static final String COMMA_VALUE = ",";
+    private static final String WHITESPACE = System.lineSeparator();
```
hronograf, line 17: is it `WHITESPACE` or it is lineSeparator? 
because we have a whitespace when we press `space` and we have line separator when we press `enter`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,83 @@
...
+
+    private void saveReport(String[] report, Path path) {
+        createFile(path);
+        for (String line : report) {
+            try {
+                Files.write(path, line.getBytes(), StandardOpenOption.WRITE);
+            } catch (IOException e) {
+                throw new RuntimeException("Can't write in file", e);
+            }
+        }
```
hronograf, line 42: with such implementation we will overwrite file each time when `Files.write` will be called
so maybe it would be better to write full text only once
try to use debugger and figure out what is going here (look at value of `line` variable at each iteration)

hronograf (changes requested):

PR #633, opened 2022-10-05
Read and rite file

Andew-Miroshnikov:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,50 @@
...
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName, true))) {
+            StringBuilder builder = new StringBuilder();
+            builder.append("supply,").append(supplyAmount).append(System.lineSeparator())
+                    .append("buy,").append(buyAmount).append(System.lineSeparator())
+                    .append("result,").append(result);
+            bufferedWriter.write(builder.toString());
+
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
```
KostyaNikiforov, line 48: Let's separate this logic into three private methods.
Please, make your read and write methods reusable. Write method must only write string data that it receive in the params

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,50 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private int supplyAmount = 0;
+    private int buyAmount = 0;
```
KostyaNikiforov, line 11: It must be method variables, but not fields

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
Perfect!):fire:

PR #634, opened 2022-10-05
created a report with given statistics

ana-rom:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,87 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Cannot write data to file", e);
+        } finally {
+            try {
+                bufferedWriter.close();
+            } catch (IOException e) {
+                throw new RuntimeException("Cannot write data to file", e);
+            }
+        }
```
ChumakAn, line 41: it's better to make an separate method with this logic and use try-with-resources

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,87 @@
...
+    }
+
+    public void readStatistic(String fromFileName) {
+        File file = new File(fromFileName);
+
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
+            String inputData = bufferedReader.readLine();
```
ChumakAn, line 48: Optional, but we can avoid creating redundant variable 
```suggestion
        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(new File(fromFileName)))) {
            String inputData = bufferedReader.readLine();
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,87 @@
...
+
+            while (inputData != null) {
+                if (inputData.startsWith(SUPPLY)) {
+                    supplyAmount = supplyAmount
+                            + Integer.parseInt(inputData.substring(inputData.indexOf(",") + 1));
```
ChumakAn, line 53: "," can be a constant

ChumakAn (changes requested):
try to divide the method into smaller methods which have only one purpose and not overcomplicate the method logic

src/main/java/core/basesyntax/WorkWithFile.java
ChumakAn: I have added an suggested piece of code

ChumakAn (approved):

PR #637, opened 2022-10-05
added logic to getStatistic mthd

Alex-Kharchuk:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,55 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
 
     public void getStatistic(String fromFileName, String toFileName) {
```
KostyaNikiforov, line 11: Let's separate your solution into three private methods.
One for reading data from file.
One for writing data to the file.
And one for statistic generation.

Such core will be more readable, reusable and easy to support

KostyaNikiforov (changes requested):
Your solution is okay.

KostyaNikiforov (approved):

PR #644, opened 2022-10-07
rewrite updated data to another file

IvanDrebot:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,57 @@
...
+        try (BufferedReader reader = new BufferedReader(new FileReader(readFile))) {
+            String value = reader.readLine();
+
+            while (value != null) {
+                builder.append(value)
+                        .append("_");
+                value = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
liliia-ponomarenko, line 25: add informative messages for exceptions

liliia-ponomarenko (changes requested):
Read common mistakes again. You need split logic on several private methods and call them in the public  ```getStatistic()``` method. Also, make all strange strings or numbers in the code constants. The name of the constant should display this object’s purpose.

liliia-ponomarenko (changes requested):
I can’t find any constants in your solution. All numbers, strings, and separators should make constants. Read an article about constant names.

hronograf (approved):
Good job!

PR #662, opened 2022-10-24
hw-work-with-file

ilko-ilya:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,63 @@
...
+                line = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file", e);
+        }
+        return builder.toString();
+    }
+
+    private String createReport(String data) {
+        String[] split1 = data.split(" ");
```
aaananas, line 34:
```suggestion
        String[] records = data.split(" ");
```
names with numbers like this are not informative

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,63 @@
...
+        int countSupply = 0;
+        int countBuy = 0;
+        for (String element : split1) {
+            String[] split = element.split(",");
```
aaananas, line 38:
```suggestion
            String[] splittedLine = element.split(",");
```

aaananas (approved):

PR #669, opened 2022-11-03
Implemented class getStatistic().Added methods readTheFile(),countSum…

AntonKalenskiy:
…() and writeTheFile() with separate logics.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,71 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int OPERATION = 0;
+    private static final int AMOUNT = 1;
+    private static final String REGEX = ",";
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private StringBuilder stringBuilder;
+    private File file;
+    private int sumSupply = 0;
+    private int sumBuy = 0;
 
     public void getStatistic(String fromFileName, String toFileName) {
+        readTheFile(fromFileName);
+        countSum();
+        writeTheFile(toFileName);
+    }
 
+    private void readTheFile(String fromFileName) {
+        file = new File(fromFileName);
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
+            stringBuilder = new StringBuilder();
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(System.lineSeparator());
+                value = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("can`t read the file" + fromFileName, e);
+        }
+    }
+
+    private void writeTheFile(String toFileName) {
+        file = new File(toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, true))) {
+            stringBuilder = new StringBuilder().append(SUPPLY).append(REGEX).append(sumSupply)
+                            .append(System.lineSeparator()).append(BUY).append(REGEX).append(sumBuy)
+                            .append(System.lineSeparator()).append(RESULT).append(REGEX)
+                            .append(Integer.valueOf(sumSupply - sumBuy));
+            bufferedWriter.write(stringBuilder.toString());
+            bufferedWriter.flush();
+        } catch (IOException e) {
+            throw new RuntimeException("can`t write to file" + toFileName, e);
+        }
+    }
+
+    private void countSum() {
+        String[] splitArray = stringBuilder.toString().split(System.lineSeparator());
+        for (String operationTypeAndValue : splitArray) {
+            String[] insideOperation = operationTypeAndValue.split(REGEX);
+            switch (insideOperation[OPERATION]) {
+                case SUPPLY:
+                    sumSupply += Integer.parseInt(insideOperation[AMOUNT]);
+                    break;
+                case BUY:
+                default:
+                    sumBuy += Integer.parseInt(insideOperation[AMOUNT]);
+                    break;
+            }
+        }
     }
 }


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,71 @@
...
+    private int sumSupply = 0;
+    private int sumBuy = 0;
 
     public void getStatistic(String fromFileName, String toFileName) {
+        readTheFile(fromFileName);
+        countSum();
+        writeTheFile(toFileName);
+    }
+
+    private void readTheFile(String fromFileName) {
```
zachar-chorny, line 28: This method should not be void. Return in it what you read from the file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,71 @@
...
+    private static final int OPERATION = 0;
+    private static final int AMOUNT = 1;
+    private static final String REGEX = ",";
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private StringBuilder stringBuilder;
+    private File file;
+    private int sumSupply = 0;
+    private int sumBuy = 0;
```
zachar-chorny, line 20: These variables should not be shared, create them in the methods where they are needed

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,71 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("can`t read the file" + fromFileName, e);
+        }
+    }
+
+    private void writeTheFile(String toFileName) {
+        file = new File(toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, true))) {
+            stringBuilder = new StringBuilder().append(SUPPLY).append(REGEX).append(sumSupply)
```
zachar-chorny, line 45: This method should only write to the file, this logic should be in the method where you generate the result

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,71 @@
...
+                            .append(System.lineSeparator()).append(RESULT).append(REGEX)
+                            .append(Integer.valueOf(sumSupply - sumBuy));
+            bufferedWriter.write(stringBuilder.toString());
+            bufferedWriter.flush();
+        } catch (IOException e) {
+            throw new RuntimeException("can`t write to file" + toFileName, e);
+        }
+    }
 
+    private void countSum() {
```
zachar-chorny, line 56:
```suggestion
    private String createReport() {
```

zachar-chorny (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
AntonKalenskiy: Why I can`t use void methods. Is it forbidden or ... I dont understand

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
 
     public void getStatistic(String fromFileName, String toFileName) {
+        StringBuilder stringBuilder = readTheFile(fromFileName);
+        String resultData = createReport(stringBuilder);
+        writeTheFile(toFileName, resultData);
+    }
+
+    private StringBuilder readTheFile(String fromFileName) {
```
```suggestion
    private String readTheFile(String fromFileName) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(System.lineSeparator());
+                value = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("can`t read the file" + fromFileName, e);
+        }
+        return stringBuilder;
```
kozhukhovsky, line 36:
```suggestion
        return stringBuilder.toString();
```

kozhukhovsky: "Why I can`t use void methods. Is it forbidden or ... I dont understand"

It can be used if this method should simply do something and you do not expect any data from it.
The task of this method is to read the file. We can’t just read, we need data from there. So we read and return data from the file to use elsewhere.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,78 @@
...
+            bufferedWriter.write(resultData);
+            bufferedWriter.flush();
+        } catch (IOException e) {
+            throw new RuntimeException("can`t write to file" + toFileName, e);
+        }
+    }
+
+    private String createReport(StringBuilder stringBuilder) {
+        int[] count = countSum(stringBuilder);
+        int sumSupply = count[0];
```
kozhukhovsky, line 51: magic number, move index 0 and 1 to the field variable, like private static final int OPERATION = 0;

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,80 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("can`t read the file" + fromFileName, e);
+        }
+        return stringBuilder.toString();
+    }
+
+    private void writeTheFile(String toFileName, String resultData) {
+        File file = new File(toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, true))) {
```
kateryna-deshchinska, line 43:
```suggestion
        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file))) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,80 @@
...
+            bufferedWriter.write(resultData);
+            bufferedWriter.flush();
```
kateryna-deshchinska, line 45:
```suggestion
```

kateryna-deshchinska (approved):

PR #698, opened 2022-11-28
hw-work-with-files-solution

Bogdan-Pokhilko:
Created constant values;
Used BufferReader to read from *.csv file, with if/else and .split() method separated and added values for supply and buy; Created the String array with the total report;
with help of try with the recourses wrote total report from String array.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,8 +1,49 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SPLIT_BY = ",";
+    private static final String SUPPLY_INDEX = "supply";
+    private static final String BUY_INDEX = "buy";
+    private static final String RESULT_INDEX = "result";
 
     public void getStatistic(String fromFileName, String toFileName) {
 
+        String valueF = "";
+        int supplyInt = 0;
+        int buyInt = 0;
+        int sum = 0;
+
+        File files = new File(toFileName);
+
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
+
+            while ((valueF = reader.readLine()) != null) {
+                if ((valueF.split(SPLIT_BY)[0]).equals(SUPPLY_INDEX)) {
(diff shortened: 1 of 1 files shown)


zachar-chorny (changes requested):
Good job. But it would be better to divide this logic into three methods :
String readFromFile(String fileName)
String crateReport(String data)
void writeReportToFile(String fileName, String report)
and call them one by one in the getStatistic method. Also create constants for the indexes by which you get data from the array

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,67 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SPLIT_BY = ",";
+    private static final String SUPPLY_INDEX = "supply";
+    private static final String BUY_INDEX = "buy";
+    private static final String RESULT_INDEX = "result";
```
aaananas, line 13: those values are not indexes

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,67 @@
...
+                .append(BUY_INDEX).append(SPLIT_BY)
+                .append(buyInt).append(System.lineSeparator())
+                .append(RESULT_INDEX).append(SPLIT_BY).append(supplyInt - buyInt);
+        String result = stringBuilder.toString();
+        return result;
+    }
 
+    public void writeReportToFile(String fileName, String report) {
+        try {
+            BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(fileName));
```
aaananas, line 60:
```suggestion
        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(fileName))) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,67 @@
...
+            throw new RuntimeException("Can't read the file", e);
+        }
+        String result = stringBuilder.toString();
+        return result;
+    }
+
+    public String createReport(String data) {
+        int supplyInt = 0;
+        int buyInt = 0;
+        int sum = 0;
```
aaananas, line 39:
```suggestion
        int supplyValue = 0;
        int buyValue = 0;
```

aaananas (approved):

aaananas (approved):
Якщо вже отримав апрув то не потрібно робити реквест знову

PR #706, opened 2022-12-02
getStatistic method was implemented

katrevichdv:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,57 @@
...
+                .append(statistics[SUPPLY_OPERATION_INDEX] - statistics[BUY_OPERATION_INDEX])
+                .toString();
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(info);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to the file " + fileName, e);
+        }
+    }
 
+    private int[] readStatistics(String fileName) {
```
kozhukhovsky, line 39: It would be better if we split this into two separate methods.
1. Reading from a data file
2. Processing this data to create a report

kozhukhovsky (approved):
Good job!

PR #711, opened 2022-12-04
Solving the jv-work-with-file task

Iskamele:
- DATA_LINE_SEPARATOR for line-by-line separation of input data into operation type and amount;
- DATA_BUILDER_SEPARATOR to separate the result of a file read and written by StringBuilder;
- OPERATION_POSITION to read the operation type in a split line;
- AMOUNT_POSITION for amount respectively.

In the getStatistic() method, a new File object is initialized, which is then passed to the readFromFile() method.

Added a readFromFile() method that reads a file using the BufferedReader and writes the result to a StringBuilder, after which it is passed to the calculateData() method before being passed to the writeToFile() method for writing.

The calculateData() method has been added, in which the supply and buy variables are declared for ease of reading. After that, a String[] array is declared, which is initialized by builder.toString(), split by DATA_BUILDER_SEPARATOR (" ", since the StringBuilder formed in the previous method contains spaces instead of a line separator). Then we cycle through the new array. Inside the cycle, we create another auxiliary array, into which we pass the element of the first array, split by DATA_LINE_SEPARATOR (","). Using if, we calculate the supply and buy values, knowing their exact indices. After the cycle, we reset the StringBuilder, after which we add the calculated data to it, according to the template:
```
supply,_amount_
buy,_amount_
result,_amount_
```

We return StringBuilder.

The writeToFile() method has been added, which, after counting, creates a new File object, initializes it with the name toFileName, and writes a StringBuilder to it byte by byte.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,11 +1,10 @@
 package core.basesyntax;
 
 import java.io.BufferedReader;
-import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
 import java.nio.file.Files;
-import java.nio.file.StandardOpenOption;
+import java.nio.file.Path;
 
 public class WorkWithFile {
     public static final String DATA_LINE_SEPARATOR = ",";
@@ -14,30 +13,31 @@ public class WorkWithFile {
     public static final int AMOUNT_POSITION = 1;
 
     public void getStatistic(String fromFileName, String toFileName) {
-        File file = new File(fromFileName);
-        readFromFile(file, toFileName);
+        String dataFromFile = readFromFile(fromFileName);
+        String report = createReport(dataFromFile);
+        writeFile(report, toFileName);
     }
 
-    private void readFromFile(File file, String toFileName) {
+    private String readFromFile(String fromFileName) {
         try {
             StringBuilder builder = new StringBuilder();
-            BufferedReader reader = new BufferedReader(new FileReader(file));
+            BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
             String value = reader.readLine();
             while (value != null) {
                 builder.append(value)
                         .append(" ");
                 value = reader.readLine();
             }
-            writeToFile(calculateData(builder), toFileName);
+            return builder.toString();
         } catch (IOException e) {
-            throw new RuntimeException("Can't read file", e);
+            throw new RuntimeException("Can't read data from file " + fromFileName, e);
         }
     }
 
-    private StringBuilder calculateData(StringBuilder builder) {
+    private String createReport(String dataFromFile) {
         int supply = 0;
         int buy = 0;
-        String[] array = builder.toString().split(DATA_BUILDER_SEPARATOR);
+        String[] array = dataFromFile.split(DATA_BUILDER_SEPARATOR);
         for (String line : array) {
             String[] splitLine = line.split(DATA_LINE_SEPARATOR);
             if (splitLine[OPERATION_POSITION].equals("supply")) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,76 @@
...
+        try {
+            StringBuilder builder = new StringBuilder();
+            BufferedReader reader = new BufferedReader(new FileReader(file));
+            String value = reader.readLine();
+            while (value != null) {
+                builder.append(value)
+                        .append(" ");
+                value = reader.readLine();
+            }
+            writeToFile(calculateData(builder), toFileName);
```
Rommelua, line 31: Method called `readFromFile` must read. It can not write

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,76 @@
...
+        return builder;
+    }
 
+    private void writeToFile(StringBuilder calculatedData, String toFileName) {
+        File file = new File(toFileName);
+        try {
+            file.createNewFile();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't create file", e);
+        }
```
Rommelua, line 68:
```suggestion
```
Files.write will create the file if needed.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,76 @@
...
+
 public class WorkWithFile {
+    public static final String DATA_LINE_SEPARATOR = ",";
+    public static final String DATA_BUILDER_SEPARATOR = " ";
+    public static final int OPERATION_POSITION = 0;
+    public static final int AMOUNT_POSITION = 1;
 
     public void getStatistic(String fromFileName, String toFileName) {
+        File file = new File(fromFileName);
+        readFromFile(file, toFileName);
```
Rommelua, line 18:
```suggestion
        String dataFromFile = readFromFile(fromFileName);
        String report = createReport(dataFromFile);
        writeFile(report, toFileName);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,76 @@
...
+            BufferedReader reader = new BufferedReader(new FileReader(file));
+            String value = reader.readLine();
+            while (value != null) {
+                builder.append(value)
+                        .append(" ");
+                value = reader.readLine();
+            }
+            writeToFile(calculateData(builder), toFileName);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
```
Rommelua, line 33: image

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,76 @@
...
+    private void writeToFile(StringBuilder calculatedData, String toFileName) {
+        File file = new File(toFileName);
+        try {
+            file.createNewFile();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't create file", e);
+        }
+        try {
+            Files.write(file.toPath(), calculatedData.toString().getBytes(),
+                    StandardOpenOption.APPEND);
```
Rommelua, line 71:
```suggestion
            Files.write(Path.of(toFileName), calculatedData.toString());
```

Rommelua (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
Iskamele: I don't fully understand this request. After all, the read data must be sent for further calculation.
I'll make a `return `in the method.

src/main/java/core/basesyntax/WorkWithFile.java
Iskamele: No, this is obviously not the case, since Files.write() requires the Path path to the created file as its first argument.
If you write the path and pass it to Files.write()
```
Path path = Paths.get(toFileName);
         try {
             Files.write(path, calculatedData.toString().getBytes(),
                     StandardOpenOption.APPEND);
         } catch (IOException e) {
             throw new RuntimeException("Can't write data to file " + toFileName, e);
         }
```
There will be an error. Show me an example of how Files.write works without first creating a new file. For the first 5-6 links in Google, there is no relevant information so that this method would work with the upper try-catch block removed.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,64 @@
...
+    public static final int AMOUNT_POSITION = 1;
 
     public void getStatistic(String fromFileName, String toFileName) {
+        String dataFromFile = readFromFile(fromFileName);
+        String report = createReport(dataFromFile);
+        writeFile(report, toFileName);
+    }
+
+    private String readFromFile(String fromFileName) {
+        try {
```
aaananas, line 22:
```suggestion
        try(BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
```

aaananas (approved):

PR #720, opened 2022-12-09
Hw workfiles

proggipg:


src/main/java/core/basesyntax/StatisticData.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class StatisticData {
```
kozhukhovsky, line 3: Great idea, but it's not necessary for this task

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,93 @@
...
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(file, true))) {
+            report.append("supply,").append(supply).append(System.lineSeparator()).append("buy,")
+                    .append(buy).append(System.lineSeparator()).append("result,").append(result);
+            writer.write(report.toString());
+        } catch (IOException e) {
+            throw new RuntimeException("IO exception", e);
+        }
+    }
+
+    private static int countLines(String fileName) {
```
kozhukhovsky, line 81: necessary method

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,93 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
 
     public void getStatistic(String fromFileName, String toFileName) {
```
kozhukhovsky, line 13: It is better to divide all the logic into three methods:
1. Reading information from a file, in what form it is, and in this way, it can be read into a string, it can be into an array of strings.
2. Information processing and report creation.
3. Writing to the file of the created report.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,93 @@
...
+        final int columnsNumber = 2;
+        final int linesNumber = countLines(fileName);
+
+        File file = new File(fileName);
+        StatisticData[] data = new StatisticData[linesNumber];
+
+        String[] lines = new String[columnsNumber];
+        int dataCounter = 0;
+
+        try {
```
kozhukhovsky, line 48: use try with resource

kozhukhovsky (changes requested):
Try to remake the solution based on the given structure.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,92 @@
...
+            throw new RuntimeException("Cannot open file", e);
+        } catch (IOException e) {
+            throw new RuntimeException("IO exception", e);
+        } catch (NullPointerException e) {
+            throw new RuntimeException("Null pointer exception");
+        }
+        return data;
+    }
+
+    private void writeToFile(int supply, int buy, int result, String fileName) {
```
aleksandr-bevza, line 68: Method "writeToFile" should not contain any logic about report, supply, buy, result. It should only write to file. 
Move all report-building logic out of this method. Please follow single responsibility principle

Make it like this:
public void writeToFile(String report, String toFileName) {
...

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,92 @@
...
 public class WorkWithFile {
 
+    private static final int columnsNumber = 2;
+
     public void getStatistic(String fromFileName, String toFileName) {
 
+        final int linesNumber = countLines(fromFileName);
+
+        String[][] operationData = new String[linesNumber][columnsNumber];
+        operationData = readFromFile(fromFileName);
```
aleksandr-bevza, line 20: You don't need to read from this file twice. No need to "count lines"
```suggestion
        String[][] operationData = readFromFile(fromFileName);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,92 @@
...
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(file, true))) {
+            report.append("supply,").append(supply).append(System.lineSeparator()).append("buy,")
+                    .append(buy).append(System.lineSeparator()).append("result,").append(result);
+            writer.write(report.toString());
+        } catch (IOException e) {
+            throw new RuntimeException("IO exception", e);
+        }
+    }
+
+    private static int countLines(String fileName) {
```
aleksandr-bevza, line 81:
```suggestion
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,92 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
 
+    private static final int columnsNumber = 2;
+
     public void getStatistic(String fromFileName, String toFileName) {
```
aleksandr-bevza, line 15: I suggest to implement this method like this:
```suggestion
    public void getStatistic(String fromFileName, String toFileName) {
        List<String> dataFromFile = readFromFile(fromFileName);
        String report = createReport(dataFromFile);
        writeToFile(report, toFileName);
}
```

aleksandr-bevza (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,8 +1,76 @@
...
+                data.append(value).append(',');
+                value = reader.readLine();
+            }
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("Cannot open file", e);
+        } catch (IOException e) {
+            throw new RuntimeException("IO exception", e);
+        } catch (NullPointerException e) {
+            throw new RuntimeException("Null pointer exception");
+        }
```
aaananas, line 36:
```suggestion
        } catch (IOException e) {
            throw new RuntimeException("Cannot open file " + fileName, e);
        }
```

PR #741, opened 2022-12-14
build success

solomiia-voloshyn:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+        writeData(report, toFileName);
+    }
+
+    private String readFile(String fromFileName) {
+        StringBuilder dataBuilder = new StringBuilder();
+        String value;
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            value = bufferedReader.readLine();
+            while (value != null) {
+                dataBuilder.append(value).append(" ");
```
KostyaNikiforov, line 25: You can try to use List instead of StringBuilder, to avoid splitting 
```suggestion
                result.add(value);
```

KostyaNikiforov: Or you can try to use Files.readAllLines method

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+        return reportBuilder.append("supply,").append(sumSupply)
+                .append(System.lineSeparator())
+                .append("buy,").append(sumBuy).append(System.lineSeparator())
+                .append("result,").append(result).toString();
+    }
 
+    public void writeData(String report, String toFileName) {
+        File resultFile = new File(toFileName);
+        try {
+            Files.write(resultFile.toPath(), report.getBytes());
```
KostyaNikiforov, line 56: Use writeString method

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+            } else {
+                sumBuy += Integer.parseInt(splitterLine[AMOUNT_INDEX]);
+            }
+        }
+        int result = sumSupply - sumBuy;
+        StringBuilder reportBuilder = new StringBuilder();
+        return reportBuilder.append("supply,").append(sumSupply)
+                .append(System.lineSeparator())
+                .append("buy,").append(sumBuy).append(System.lineSeparator())
+                .append("result,").append(result).toString();
```
KostyaNikiforov, line 50: (Optional) Actually, you can use ordinary concatenation instead of StringBuilder in such case
```suggestion
        return new StringBuilder().append("supply,").append(sumSupply)
                .append(System.lineSeparator())
                .append("buy,").append(sumBuy).append(System.lineSeparator())
                .append("result,").append(result).toString();
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #742, opened 2022-12-14
Jv work with file hw-1st commit

OleksandrKostiukevych:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        String stringsFromFile = readFromFile(fromFileName);
+        String report = createReport(stringsFromFile);
+        writeReport(report, toFileName);
+    }
+
+    private String readFromFile(String fileName) {
+        StringBuilder stringBuilder = new StringBuilder();
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName));
```
KostyaNikiforov, line 24: Why don't you use try-with-resources?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(';');
```
KostyaNikiforov, line 28: You can try to use LIst instead of the Builder, to avoid splitting
```suggestion
                result.add(value;)
```

KostyaNikiforov: Also, you can try to use Filed.readAllLines method

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+    private String createReport(String readFromFile) {
+        StringBuilder stringBuilder = new StringBuilder();
+        String[] splitter = readFromFile.split(";");
+        int supplySum = 0;
+        int buySum = 0;
+        for (String element : splitter) {
+            supplySum += element.charAt(0) == FIRST_PARAMETER.charAt(0) ? Integer.parseInt(element
+                    .substring(FIRST_PARAMETER.length() + 1)) : 0;
+            buySum += element.charAt(0) == SECOND_PARAMETER.charAt(0) ? Integer.parseInt(element
+                    .substring(SECOND_PARAMETER.length() + 1)) : 0;
```
KostyaNikiforov, line 46: Do you really need to try both cases? Maybe it is better to use if-else construction here?

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
OleksandrKostiukevych: Wanted to do so at first, but were broblems. Do not understand, where is a mistake. Help please!
https://prnt.sc/Iu-qpp9CYuW2

KostyaNikiforov (approved):

PR #775, opened 2023-01-09
Added logic to method getStatistic

ivbudenkov:
Added methods and constants for reading and writing data.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,70 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 
+public class WorkWithFile {
+    public static final String SUPPLY_FIELD = "supply";
+    public static final String BUY_FIELD = "buy";
+    public static final String RESULT_FIELD = "result";
+    public static final int NAME_INDEX = 0;
+    public static final int VALUE_INDEX = 1;
+    public static final String DELIMITER = ",";
+
+    public void getStatistic(String fromFileName, String toFileName) {
+        Map<String, Integer> data = readDataFromFile(fromFileName);
+        String report = createReport(data);
+        writeReportToFile(report, toFileName);
+    }
+
+    private void writeReportToFile(String report, String toFileName) {
+        File file = new File(toFileName);
+        try {
+            file.createNewFile();
+            BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file));
+            bufferedWriter.write(report.toString());
+            bufferedWriter.close();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to the file " + toFileName, e);
+        }
+    }
+
+    private String createReport(Map<String, Integer> data) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+        File file = new File(fromFileName);
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
+            String line = bufferedReader.readLine();
+            while (line != null) {
+                String[] values = line.split(DELIMITER);
+                String key = values[NAME_INDEX];
+                result.put(key,result.get(key) + Integer.valueOf(values[VALUE_INDEX]));
+                line = bufferedReader.readLine();
+            }
+        } catch (FileNotFoundException e) {
```
nacenik, line 63: You have the same logic in two catch blocks, so delete this one. You can use only IOException in the catch, cause FileNotFoundException is a sub-class of IOException.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
+        Map<String, Integer> data = readDataFromFile(fromFileName);
+        String report = createReport(data);
+        writeReportToFile(report, toFileName);
+    }
+
+    private void writeReportToFile(String report, String toFileName) {
+        File file = new File(toFileName);
+        try {
```
nacenik, line 29: [Use try-with-resourses here](https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file#try-with-resources)
Here is enough to make something like 
```
try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(fileName))) {
         bufferedWriter.write(report);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+            file.createNewFile();
+            BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file));
+            bufferedWriter.write(report.toString());
```
nacenik, line 32: you don't need to call toString() method in string object

nacenik (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,68 @@
...
+    }
+
+    private void writeReportToFile(String report, String toFileName) {
+        File file = new File(toFileName);
+        try {
+            file.createNewFile();
+            try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file))) {
+                file.createNewFile();
+                bufferedWriter.write(report);
+            }
```
aaananas, line 33: dont add try inside a try block

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,68 @@
...
+        stringBuilder.append(SUPPLY_FIELD).append(DELIMITER).append(data.get(SUPPLY_FIELD))
+                .append(System.lineSeparator())
+                .append(BUY_FIELD).append(DELIMITER).append(data.get(BUY_FIELD))
+                .append(System.lineSeparator())
+                .append(RESULT_FIELD).append(DELIMITER)
+                .append(data.get(SUPPLY_FIELD) - data.get(BUY_FIELD));
+        return stringBuilder.toString();
+    }
 
+    private Map<String, Integer> readDataFromFile(String fromFileName) {
```
aaananas, line 50: lets make this method responsible for just 1 task - read data from file
parse data in another method
```suggestion
    private String readDataFromFile(String fromFileName) {
```

aaananas (changes requested):

PR #777, opened 2023-01-10
Task is done.

Viper7557:


src/main/java/File.java
```diff
@@ -0,0 +1,8 @@
+import core.basesyntax.WorkWithFile;
+
+public class File {
+    public static void file(String[] args) {
+        WorkWithFile workWithFile = new WorkWithFile();
+        workWithFile.getStatistic("apple.csv","result.csv");
+    }
```
aaananas, line 7: remove this class

aaananas (changes requested):
run `mvn clean package`

PR #790, opened 2023-01-18
getStatistic() method was implemented

MaxSkory:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,52 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        StringBuilder report = createReport(fromFileName);
+        writeToFile(report, toFileName);
+    }
+
+    private void writeToFile(StringBuilder report, String toFileName) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName))) {
+            writer.write(report.toString());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to file" + e);
```
Ivan95kos, line 20: Write a more informative message, and the same for the second exception
```suggestion
            throw new RuntimeException("Can't write data to the file " + toFileName, e);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,52 @@
...
+        }
+    }
 
+    private StringBuilder createReport(String fromFileName) {
```
Ivan95kos, line 24: This method reads from a file and also generates a report.
Divide program logic

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,52 @@
...
+                if (line == null) {
+                    break;
+                } else if (line.contains("supply")) {
+                    value = line.split(",")[1];
+                    supplyTotal += Integer.parseInt(value);
+                } else {
+                    value = line.split(",")[1];
+                    buyTotal += Integer.parseInt(value);
+                }
+            }
```
Ivan95kos, line 42: Perpetual loop is bad practice.
It is better to set a condition for better readability
```suggestion
            line = reader.readLine();
            while (line != null) {
                if (line.contains("supply")) {
                    value = line.split(",")[1];
                    supplyTotal += Integer.parseInt(value);
                } else {
                    value = line.split(",")[1];
                    buyTotal += Integer.parseInt(value);
                }
                line = reader.readLine();
            }
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+
```
okuzan, line 10: delet this line

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+        secondTotal = 0;
+        readFile(fromFileName);
+        createReport();
+        writeToFile(toFileName);
+    }
+
+    private void readFile(String fromFileName) {
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
+            String line;
+            line = reader.readLine();
```
okuzan, line 30:
```suggestion
            String line = reader.readLine();
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+            while (line != null) {
+                calculateTotal(line);
+                line = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file" + e);
```
okuzan, line 36: add filename

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+import java.io.IOException;
+
 public class WorkWithFile {
+
+    private static final String[] OPERATION_NAME = {"supply", "buy"};
+    private static final String DELIMITER = ",";
+    private static final String RESULT_NAME = "result";
+    private StringBuilder report;
+    private int firstTotal;
+    private int secondTotal;
```
okuzan, line 16: not informative naming, how to guess what it stores?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+
+    private static final String[] OPERATION_NAME = {"supply", "buy"};
```
okuzan, line 11: better create separate supply operation constant and buy operation constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+    private static final String DELIMITER = ",";
+    private static final String RESULT_NAME = "result";
+    private StringBuilder report;
```
okuzan, line 14: we don't need it as field, just as a local variable used to return String

okuzan (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY_LINE_NAME = "supply";
+    private static final String BUY_LINE_NAME = "buy";
+    private static final String RESULT_LINE_NAME = "result";
+    private static final String DELIMITER = ",";
+    private int supplyTotal;
+    private int buyTotal;
```
aaananas, line 15: those should be local variables

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
+        supplyTotal = 0;
+        buyTotal = 0;
+        readFile(fromFileName);
+        writeToFile(createReport(), toFileName);
```
aaananas, line 21:
```suggestion
        String dataFromFile = readFile(fromFileName);
        String report = createReport(dataFromFile);
        writeToFile(report, toFileName);
```
your program should look like this

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+                line = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file" + fromFileName, e);
+        }
+    }
+
+    private void calculateTotal(String line) {
+        String amount;
+        if (line.contains(SUPPLY_LINE_NAME)) {
```
aaananas, line 38: better use split(",") method and compare the exact line with equals

aaananas (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
MaxSkory: why?

andriy-shulepa (approved):

PR #804, opened 2023-02-02
work with CVS statistic

kromch:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,47 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
-
+        String line;
+        HashMap<String, Integer> st = new HashMap<>();
+        try (BufferedReader bri = new BufferedReader(new FileReader(fromFileName));
+                BufferedWriter bro = new BufferedWriter(new FileWriter(toFileName))) {
```
petro-ushchuk, line 15: “write to file” or “create report” - are a separate operations. So you should create separate methods for them.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,47 @@
...
+            while ((line = bri.readLine()) != null) {
+                String[] em = line.split(",");
+                if (em.length == 2) {
+                    if (st.containsKey(em[0].trim())) {
+                        st.replace(em[0].trim(), st.get(em[0].trim()) + Integer.parseInt(em[1]
+                                .trim()));
+                    } else {
+                        st.put(em[0].trim(), Integer.parseInt(em[1].trim()));
+                    }
```
petro-ushchuk, line 24: You call em[0].trim() 4 times and em[1].trim() 2 times. Better to introduce it as local variable.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,47 @@
...
+                    } else {
+                        dif -= st.get(key);
+                    }
+                    i++;
+                }
+                bro.write(tmp);
+                bro.write(String.format("result,%d\n", Math.abs(dif)));
+            }
+        } catch (IOException | NumberFormatException e) {
+            e.printStackTrace();
```
petro-ushchuk, line 44: Stack trace is printed by default, it is bad practice. Please, re-throw RuntimeException with original exception and some message in the parameters.

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashMap;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
-
+        String line;
```
petro-ushchuk, line 12: Better move this variable inside `try` block

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+        HashMap<String, Integer> st = new HashMap<>();
+        try (BufferedReader bri = new BufferedReader(new FileReader(fromFileName))) {
+            while ((line = bri.readLine()) != null) {
```
petro-ushchuk, line 15: Introduce `String line = bri.readLine()` before `while` block

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+                    }
+                }
+            }
+        } catch (IOException | NumberFormatException e) {
+            if (e instanceof IOException) {
+                throw new RuntimeException(e.getMessage() + "Problem with input file");
+            } else {
+                throw new RuntimeException(e.getMessage());
+            }
+        }
```
petro-ushchuk, line 33: Make 2 `catch` blocks instead

petro-ushchuk, line 33: Divide program logic

read and write - are a separate operations. So we should create separate methods for them.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+                bro.write(tmp);
+                bro.write(String.format("result,%d\n", Math.abs(dif)));
+            }
+        } catch (IOException | NumberFormatException e) {
+            if (e instanceof IOException) {
+                throw new RuntimeException(e.getMessage() + "Problem with output file");
+            } else {
+                throw new RuntimeException(e.getMessage());
+            }
+        }
```
petro-ushchuk, line 57: Make 2 `catch` blocks instead

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+import java.util.HashMap;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
-
+        String line;
+        HashMap<String, Integer> st = new HashMap<>();
+        try (BufferedReader bri = new BufferedReader(new FileReader(fromFileName))) {
+            while ((line = bri.readLine()) != null) {
+                String[] em = line.split(",");
```
petro-ushchuk, line 16: `","` should be constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+                if (em.length == 2) {
```
petro-ushchuk, line 17: `2` should be constant

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+                for (String key : st.keySet()) {
+                    tmp = String.format("%s,%d\n", key, st.get(key)) + tmp;
+                    if (i == 0) {
+                        dif = st.get(key);
+                    } else {
+                        dif -= st.get(key);
+                    }
+                    i++;
+                }
+                bro.write(tmp);
```
petro-ushchuk, line 48: Use `StringBuilder` instead.

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
kromch: i did better

src/main/java/core/basesyntax/WorkWithFile.java
kromch: done

src/main/java/core/basesyntax/WorkWithFile.java
kromch: done

src/main/java/core/basesyntax/WorkWithFile.java
kromch: done

src/main/java/core/basesyntax/WorkWithFile.java
kromch: refactored

src/main/java/core/basesyntax/WorkWithFile.java
kromch: i did better

src/main/java/core/basesyntax/WorkWithFile.java
kromch: done

src/main/java/core/basesyntax/WorkWithFile.java
kromch: done

PR #811, opened 2023-02-09
jv-work-with-file solution

potrap:


maxline (changes requested):
Good try! Looks like the solution is complicated a bit for this simple task. Using HashMap and sorting is not necessary here.  Let's try to implement in more simple way? Anyway it is a good attempt for the solution.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,87 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        File toFile = new File(toFileName);
+        File fromFile = new File(fromFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFile));
+                BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFile))) {
+            String resultList = getResultText(bufferedReader);
+            bufferedWriter.write(resultList);
+        } catch (IOException e) {
+            throw new RuntimeException("wrong file path");
+        }
```
nacenik, line 29: I'd recommend segregating this block into two separate methods: the first one is reading and returning result, next one is writing this result into a file. The main reason for this approvement is to make debugging easier, cause you can customize exceptions for reading and writing (modify message to add file path)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,87 @@
...
+    private static final int RESULT_INDEX = 2;
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT_VALUE = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        File toFile = new File(toFileName);
+        File fromFile = new File(fromFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFile));
+                BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFile))) {
```
nacenik, line 24: it's necessary to create File objects 
```suggestion
        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName));
                BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
```

nacenik (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,97 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        String statistic = readText(fromFileName);
+        writeText(toFileName, statistic);
+    }
+
+    private void writeText(String toFileName, String statistic) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(statistic);
+        } catch (IOException e) {
+            throw new RuntimeException("can't write to file");
```
karma-o, line 29: add fileName to the message and exception that is thrown

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,97 @@
...
+        }
+    }
+
+    private String readText(String fromFileName) {
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            return getResultText(bufferedReader);
+        } catch (IOException e) {
+            throw new RuntimeException("wrong file path");
```
karma-o, line 37: add fileName to the message and exception that is thrown

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,97 @@
...
+            if (i != 0) {
+                list.append(System.lineSeparator());
+            }
+            list.append(information[i][WORD_INDEX]).append(",")
+                    .append(information[i][VALUE_INDEX]);
+        }
+        return list.toString();
+    }
+
+    private String[][] getMapInformation(BufferedReader bufferedReader)
```
karma-o, line 54: we should not have methods that are receiving the reader as a parameter, and as it is written in the common mistakes we should close the reader/writer resources as soon as possible, so i n general I would break down our logic into 3 methods:
* String readFile(String fromFileName)
* String createReport(String data)
* void writeToFile (String report)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,97 @@
...
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(statistic);
+        } catch (IOException e) {
+            throw new RuntimeException("can't write to file");
+        }
+    }
+
+    private String readText(String fromFileName) {
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            return getResultText(bufferedReader);
```
karma-o, line 35: This method should be responsible only for reading the data and collecting it into a single place, but you are effectively calling 3 methods inside this try block inside each other. What we should do is minimize the code and logic amount inside the try block, and separate the report generation functionality from the file reading one

karma-o (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,95 @@
...
+
+    private void writeText(String toFileName, String statistic) {
+        try {
+            Files.writeString(new File(toFileName).toPath(), statistic);
+        } catch (IOException e) {
+            throw new RuntimeException("can't write to file: " + toFileName, e);
+        }
+    }
+
+    private String readText(String fromFileName) {
```
```suggestion
    private List<String> readText(String fromFileName) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,95 @@
...
+    private static final int SUPPLY_INDEX = 0;
+    private static final int BUY_INDEX = 1;
+    private static final int RESULT_INDEX = 2;
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT_VALUE = "result";
+    private static final String ZERO_VAR = "0";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String statistic = readText(fromFileName);
```
KseniiaMakarova, line 20:
```suggestion
        String statistic = getResultText(readText(fromFileName));
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,95 @@
...
+        List<String> list;
+        try {
+            list = Files.readAllLines(new File(fromFileName).toPath());
+        } catch (IOException e) {
+            throw new RuntimeException("wrong file path: " + fromFileName, e);
+        }
+        return getResultText(list);
+    }
+
+    private String getResultText(List<String> list) {
```
KseniiaMakarova, line 42: here let's just use for-each loop
introduce variables totalSupply and totalBuy

for each line, split line by comma, and according to the operation - add value either to totalSupply, or to totalBuy

then use StringBuilder to create a ready String report

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,95 @@
...
+            if (i != 0) {
+                stringList.append(System.lineSeparator());
+            }
+            stringList.append(information[i][WORD_INDEX]).append(",")
+                    .append(information[i][VALUE_INDEX]);
+        }
+        return stringList.toString();
+    }
+
+    private String[][] getMapInformation(List<String> list) {
```
KseniiaMakarova, line 55: these methods can be removed

KseniiaMakarova (changes requested):

sarakhmen (approved):
Good job :)

PR #814, opened 2023-02-10
Methods for get statistic from file, and writing in file was implement

GalubaGSV:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final String CHAR_SEPARATOR = ",";
+
+    private File myFile = null;
```
petro-ushchuk, line 16: Introduce as local variable

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+                tempStr = value.split(CHAR_SEPARATOR);
+                if (tempStr[0].equals(SUPPLY)) {
+                    supplySum += Integer.parseInt(tempStr[1]);
+                } else if (tempStr[0].equals(BUY)) {
+                    buySum += Integer.parseInt(tempStr[1]);
+                } else {
+                    throw new RuntimeException("Error in csv file");
+                }
+                value = bufferedReader.readLine();
+            }
```
petro-ushchuk, line 39: Extract `read` operation in separate method, calculations can leave here or also extract to separate method. Don't use global variables to keep data, return as result of calling method.

petro-ushchuk (changes requested):

Andriy-dz (approved):

PR #817, opened 2023-02-10
I read data from file input, create a report and write it to newFile

NazarZva:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+        result.append("result,").append(arr[0] - arr[1]);
+        text[2] = result.toString();
+        return text;
+    }
+
+    private void putDataToFile(String[] text, String putFile) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(putFile))) {
+            /*File file = new File(putFile);
+            FileWriter fileWriter = new FileWriter(file, true);
+            BufferedWriter bufferedWriter = new BufferedWriter(fileWriter);*/
```
KseniiaMakarova, line 72: remove commented code

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] text = openFile(fromFileName);
+        int[] data = calcutData(text);
+        String[] dataForPrint = printData(data);
+        putDataToFile(dataForPrint,toFileName);
+    }
+
+    private String[] openFile(String fileName) {
```
KseniiaMakarova, line 18: bad method name

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+        StringBuilder string = new StringBuilder();
+        try {
+            File file = new File(fileName);
+            FileReader fileReader = new FileReader(file);
+            BufferedReader bufferedReader = new BufferedReader(fileReader);
```
KseniiaMakarova, line 23: do you close resources?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+            String text = bufferedReader.readLine();
+            while (text != null) {
+                string.append(text).append("///");
```
KseniiaMakarova, line 26: this delimiter should be a constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+                text = bufferedReader.readLine();
+            }
+            bufferedReader.close();
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t open file!", e);
```
KseniiaMakarova, line 31: add fileName

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+        }
+        if (string.toString().isEmpty()) {
+            String[] empty = {};
+            return empty;
+        } else {
```
KseniiaMakarova, line 36: do we need else after return?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+            return string.toString().split("///");
+        }
+    }
+
+    private int[] calcutData(String[] str) {
```
KseniiaMakarova, line 41: bad method name

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+        int[] result = new int[]{0,0};
```
KseniiaMakarova, line 42:
```suggestion
        int[] result = new int[]{0, 0};
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+        for (int i = 0; i < str.length; i++) {
+            String[] strings = str[i].split(",");
+            if (strings[0].equals("supply")) {
+                result[0] += Integer.parseInt(strings[1]);
```
KseniiaMakarova, line 46: 0 and 1 should be made constants because it should be more clear what is going on

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+            } else if (strings[0].equals("buy")) {
+                result[1] += Integer.parseInt(strings[1]);
+            }
+        }
+        return result;
+    }
+
+    private String[] printData(int[] arr) {
```
KseniiaMakarova, line 54:
```suggestion
    private String createReport(int[] data) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+        String[] text = new String[3];
```
KseniiaMakarova, line 55: do you really need to create an array?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+        result.setLength(0);
+        result.append("buy,").append(arr[1]);
+        text[1] = result.toString();
+        result.setLength(0);
+        result.append("result,").append(arr[0] - arr[1]);
+        text[2] = result.toString();
+        return text;
+    }
+
+    private void putDataToFile(String[] text, String putFile) {
```
KseniiaMakarova, line 68:
```suggestion
    private void putDataToFile(String text, String putFile) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(putFile))) {
+            /*File file = new File(putFile);
+            FileWriter fileWriter = new FileWriter(file, true);
+            BufferedWriter bufferedWriter = new BufferedWriter(fileWriter);*/
 
+            for (int i = 0; i < text.length; i++) {
+                bufferedWriter.write(text[i]);
+                bufferedWriter.write("\n");
```
KseniiaMakarova, line 76: System.lineSeparator() when creating a report

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+            }
+            /*bufferedWriter.close();*/
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t put data to file!",e);
```
KseniiaMakarova, line 80: add fileName

KseniiaMakarova (changes requested):
please be more attentive to naming and code style

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] text = readFile(fromFileName);
+        int[] data = calculateDataForReport(text);
+        String dataForPrint = createReport(data);
+        putDataToFile(dataForPrint,toFileName);
+    }
+
+    private String[] readFile(String fileName) {
+        StringBuilder string = new StringBuilder();
```
karma-o, line 23: I wouldn't name `StringBuilder` as a `string`

src/main/java/core/basesyntax/WorkWithFile.java
karma-o: not fixed, we should either use try with resources or finally block for this
check [this](https://stackoverflow.com/questions/56151113/use-try-with-resources-or-close-this-bufferedreader-in-a-finally-clause) for details

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+            }
+            bufferedReader.close();
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t open file!" + fileName, e);
+        }
+        if (string.toString().isEmpty()) {
+            String[] empty = {};
+            return empty;
+        }
+        return string.toString().split(DELIMITER);
```
karma-o, line 41: this is strange to build every data line into a single string and then split it again to return an array, you could have just collected it to an array right away, or returned the whole string in this method

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String DELIMITER = "///";
+    private static final int INDEX_0 = 0;
+    private static final int INDEX_1 = 1;
```
karma-o, line 13: constant names should represent the conceptual point of view and what does it mean for us and not just a fancy number name
If you want to reuse some constant value with different ideas, it is better to create another one with the same value and name it right
```suggestion
    private static final int OPERATION_INDEX = 0;
    private static final int AMOUNT_INDEX = 1;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+            throw new RuntimeException("Can`t open file!" + fileName, e);
+        }
+        if (string.toString().isEmpty()) {
+            String[] empty = {};
+            return empty;
+        }
+        return string.toString().split(DELIMITER);
+    }
+
+    private int[] calculateDataForReport(String[] str) {
```
karma-o, line 44: not a mistake, but it is not obvious what our method returns so I would add some [Javadoc](https://www.tutorialspoint.com/java/java_documentation.htm) on what is this array and what is what by the index of that array element

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+            if (strings[INDEX_0].equals("supply")) {
+                result[INDEX_0] += Integer.parseInt(strings[INDEX_1]);
+            } else if (strings[INDEX_0].equals("buy")) {
+                result[INDEX_1] += Integer.parseInt(strings[INDEX_1]);
+            }
+        }
+        return result;
+    }
+
+    private String createReport(int[] data) {
```
karma-o, line 57: In addition to the previous comment: we can also refactor our code a bit to have `calculateDataForReport()` be called inside `createReport()` because I think they are working close enough to have it as 1 or at least to call each other

karma-o (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+                stringBuilder.append(text).append(DELIMITER);
+                text = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t open file!" + fileName, e);
+        }
+        if (stringBuilder.toString().isEmpty()) {
+            String empty = new String();
+            return empty;
+        }
```
KseniiaMakarova, line 36: why do we need this?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+        return stringBuilder.toString();
+    }
+
+    /* methods calculateDataForReport return int array with two elements.
+    * first element mean sum of all values with supply
+    * second element mean sum of all values with buy */
+    private int[] calculateDataForReport(String[] str) {
```
KseniiaMakarova, line 43:
```suggestion
    private int[] calculateDataForReport(String[] array) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+        int[] result = new int[]{0, 0};
+        for (int i = 0; i < str.length; i++) {
```
KseniiaMakarova, line 45: can we use for-each?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+                result[POSITION_OF_BUY_VALUE] += Integer.parseInt(strings[POSITION_OF_VALUE]);
+            }
+        }
+        return result;
+    }
+
+    private String createReport(String string) {
+        int[] data = calculateDataForReport(string.split(DELIMITER));
+        StringBuilder result = new StringBuilder();
+        result.append("supply,").append(data[0]).append("\n");
```
KseniiaMakarova, line 59: use your constant for 0 and System.lineSeparator() instead of \n

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+        }
+        return stringBuilder.toString();
+    }
+
+    /* methods calculateDataForReport return int array with two elements.
+    * first element mean sum of all values with supply
+    * second element mean sum of all values with buy */
+    private int[] calculateDataForReport(String[] array) {
+        int[] result = new int[]{0, 0};
+        for (String arr:array) {
```
KseniiaMakarova, line 41:
```suggestion
        for (String line : array) {
```
bad naming

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #830, opened 2023-02-13
Hw-with-file

KosMarch:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+
+    private void writeFile(int buySum, int supplySum, int result, File toFile) {
+
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFile))) {
+
+            StringBuilder stringBuilder = new StringBuilder();
+
+            stringBuilder.append("supply,").append(supplySum).append(System.lineSeparator())
+                    .append("buy,").append(buySum).append(System.lineSeparator())
+                    .append("result,").append(result);
```
maxline, line 49: exctract a separate method for these lines, e.g. `createResultText()`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int NAME_POS = 0;
+    private static final int NUM_POS = 1;
+    private int buySum;
+    private int supplySum;
+    private int result;
```
maxline, line 15: looks like it is not needed to create these variables on class level, move its to the methods where its called.

maxline (changes requested):
Good job! Let's improve it :)

KseniiaMakarova (changes requested):
1 small comment left

KseniiaMakarova (approved):

PR #851, opened 2023-02-14
Work with file

YaroslavKopiyka:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,54 @@
...
+                        break;
+                    case SUPPLY:
+                        supply = supply + Integer.parseInt(value[FIRST_INDEX]);
+                        break;
+                    default:
+                        System.err.println("Corrupted data in input file");
+                }
+                line = reader.readLine();
+            }
+        } catch (IOException e) {
```
petro-ushchuk, line 38: Extract calculation and read logic to the separate methods.

petro-ushchuk (changes requested):

petro-ushchuk (approved):
Good job!

PR #864, opened 2023-02-15
Work with file homework

yuriy49:


maxline (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+
+    private void writeReportToFile(List<String> linetFromFile, String toFileName) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(createReport(linetFromFile));
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t write data");
+        }
+    }
 
+    private String createResultString(int res, int summSuplie, int summBye) {
```
maxline, line 56: please do not cut the words `result` instead of `res`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+                summSuplie += Integer.parseInt(lineArray[LINE_VALUE]);
+            } else if (lineArray[LINE_NAME].equals(BUY)) {
+                summBye += Integer.parseInt(lineArray[LINE_VALUE]);
+            }
+        }
+        int res = summSuplie - summBye;
+        return createResultString(res, summSuplie, summBye);
+    }
+
+    private void writeReportToFile(List<String> linetFromFile, String toFileName) {
```
maxline, line 48: instead of `List<String> linetFromFile` parameter you should have here just `String report` that you get from createReport() method.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(createReport(linetFromFile));
```
maxline, line 50: call `createReport()` not inside `writeReportToFile()`, but before it.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
 public class WorkWithFile {
+    private static final String SEPARATOR = ",";
+    private static final int LINE_NAME = 0;
+    private static final int LINE_VALUE = 1;
+    private static final String BUY = "buy";
+    private static final String SUPPLY = "supply";
+    private static final String RESULT = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        writeReportToFile(readFile(fromFileName), toFileName);
```
maxline, line 19: order of methods calls should be the next:
```
 String report createReport(readFile(fromFileName);
 ```

**Review by maxline — CHANGES_REQUESTED** [review]
Good try! Let's improve solution one more time

File: src/main/java/core/basesyntax/WorkWithFile.java [hunk 5]
```diff
@@ -1,7 +1,62 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
+        setReportToFile(fromFileName,toFileName);
+    }
+
+    private List<String> getValueFromFile(String fromFileName) {
+        File fileFrom = new File(fromFileName);
+        List<String> linetFromFile;
+        try {
+            linetFromFile = Files.readAllLines(fileFrom.toPath());
```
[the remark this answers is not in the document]
karma-o: Still not fixed... please check all of the other comments regarding that
```suggestion
            linesFromFile = Files.readAllLines(fileFrom.toPath());
```

src/main/java/core/basesyntax/WorkWithFile.java
karma-o: not fixed, this is important so let's not inline the operations and do this with variables, this will be easier to understand
```suggestion
        List<String> data = readFile(fromFileName);
        String report = createReport(data);
        writeToFile(toFileName, report);
```
We are doing this because of the single responsibility that each method even a class should have
\+ better to use abstract names, since our writeToFile method can write not only some specific reports but any string actually

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+                .append(RESULT).append(SEPARATOR).append(res);
+        return report.toString();
+    }
 
+    private void setReportToFile(String fromFileName, String toFileName) {
+        File fileTo = new File(toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(getRepor(fromFileName));
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t write data");
```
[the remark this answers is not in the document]
karma-o: and a fileName \+ it is still not fixed..
Please be attentive to the reviews, and do not leave any comment without a fix or an explanation of why you did not do this!

src/test/java/core/basesyntax/WorkWithFileTest.java
```diff
@@ -8,7 +8,7 @@
 import java.nio.file.Path;
 
 public class WorkWithFileTest {
```
karma-o, line 10: would be nice to revert these changes on the tests

karma-o (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
yuriy49: I don`t realy understand with what purpose create String if we can get resulr right from method?

KseniiaMakarova (changes requested):
please be more attentive to comments
it is important that you fix all comments before requesting another review

sometimes some of the comments are hidden , it is important to see them as well

if you cannot understand a comment - you should write to Slack

KseniiaMakarova (approved):

PR #872, opened 2023-02-18
Work-with-file hw 1

vojtcekhovski:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,54 @@
...
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String SEPARATOR = ",";
+    private static final int INDEX_NAME = 0;
+    private static final int INDEX_AMOUNT = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
```
okuzan, line 17: getStatistic should call three private methods, each dedicated to one job only, 1 - read from file, 2 - create report from that info, 3 - write report  to the file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,54 @@
...
+            result = supplyCount - buyCount;
+
+            stringBuilder.append("supply,").append(supplyCount).append(System.lineSeparator())
+                    .append("buy,").append(buyCount).append(System.lineSeparator())
+                    .append("result,").append(result);
 
+            try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFile))) {
+                bufferedWriter.write(stringBuilder.toString());
+            } catch (IOException e) {
+                throw new RuntimeException("Can't write to file", e);
```
okuzan, line 48: add filename to error here and below

okuzan (changes requested):
Good job with constants :)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String SEPARATOR = ",";
+    private static final int INDEX_NAME = 0;
+    private static final int INDEX_AMOUNT = 1;
```
olekskov, line 15: Better to use naming according to general reading habits. INDEX_NAME and INDEX_AMOUNT looks more like "name of index" and "amount of index" as INDEX is first so it's main.
NAME_INDEX and AMOUNT_INDEX are more understandable

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+        int supplyCount = 0;
+        int buyCount = 0;
+        int result;
+        String[] dataArray = data.split(System.lineSeparator());
+        StringBuilder stringBuilder = new StringBuilder();
+        for (String line : dataArray) {
+            String[] lineArray = line.split(SEPARATOR);
+            if (lineArray[INDEX_NAME].equals(SUPPLY)) {
+                supplyCount += Integer.parseInt(lineArray[INDEX_AMOUNT]);
+            } else if (lineArray[INDEX_NAME].equals(BUY)) {
```
olekskov, line 52: We have binary choice here "BUY" or "SUPPLY" so not really sure that second condition check is necessary. Or if we expect something else may come - we would need to include logic for unexpected inputs

olekskov (approved):
Looks good to me. Please pay attention to small remarks for further consideration

PR #885, opened 2023-03-01
Main

OlegMR0:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SPECIAL_DELIMITER = "!Delimiter@";
+
     public void getStatistic(String fromFileName, String toFileName) {
```
kozhukhovsky, line 13: It is better to break down the logic of the program 3 methods:
1. Reading from file
2. Report creation
3. Writing the report to a file

kozhukhovsky: There should only be logic in this method:

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,83 @@
...
+        File readFromFile = new File(fromFileName);
+        File writeToFile = new File(toFileName);
+        String[] fileArray = simplificateArray(convertFileToStringArray(readFromFile));
+        if (!fileArray[0].contains("supply")) {
```
kozhukhovsky, line 17: move index to the class constant

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
uliana-andrushchuk: not fixed. please move all strings like `",%d"`, `','` to constants

uliana-andrushchuk (changes requested):

kozhukhovsky (approved):

PR #896, opened 2023-03-13
hw-jv-work-with-file

andrii-subash:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int SUPPLY_OR_BUY = 0;
```
okuzan, line 11:
```suggestion
    private static final int OPERATION_INDEX = 0;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+    private static final int VALUE = 1;
```
okuzan, line 12:
```suggestion
    private static final int AMOUNT_INDEX = 1;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+    private static final String SEPARATOR = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
```
okuzan, line 15: [too much logic ](https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file#divide-program-logic)in a single method. Decompose the task: `getStatistic` should have 3 method calls, responsible for: retrieving data, working with it (generating result), and writing result to a file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+            newFile.createNewFile();
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t create file", e);
+        }
+        fileWriter(newFile, stringBuilder.toString());
+    }
+
+    private String fileReader(File file) {
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
```
okuzan, line 51: https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file#try-with-resources

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+            StringBuilder stringBuilder = new StringBuilder();
+            String value = reader.readLine();
+
+            while (value != null) {
+                stringBuilder.append(value).append(" ");
+                value = reader.readLine();
+            }
+            return stringBuilder.toString();
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t read file", e);
```
okuzan, line 61: include filename in the message

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+    }
+
+    private String fileReader(File file) {
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(file));
+            StringBuilder stringBuilder = new StringBuilder();
+            String value = reader.readLine();
+
+            while (value != null) {
+                stringBuilder.append(value).append(" ");
```
okuzan, line 56: it's a bad choice for a custom separator. What if your separator will change to " " instead of a current ","? it will break your logic. go with System.lineSeparator

okuzan (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+                .append("result").append(SEPARATOR).append(supply - buy);
+
+        return stringBuilder.toString();
+    }
 
+    private void fileWriter(File file, String text) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file))) {
+            bufferedWriter.write(text);
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t write data to file", e);
```
okuzan, line 73: include filename or path in the message

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+
+        try {
+            newFile.createNewFile();
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t create file", e);
+        }
+        fileWriter(newFile, getResult(readFile));
+    }
+
+    private String fileReader(File file, String fileName) {
```
okuzan, line 28: noun is a bad choice for a method name

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int OPERATION_INDEX = 0;
+    private static final int AMOUNT_INDEX = 1;
+    private static final String SEPARATOR = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        File file = new File(fromFileName);
+        String readFile = fileReader(file, fromFileName);
```
okuzan, line 17:
```suggestion
        String content = readFile(fromFileName);
```

src/main/java/core/basesyntax/WorkWithFile.java
okuzan, line 28: method required redundant info, both file and fileName, you can get one from the other. Delete `file` from the arguments

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+        File newFile = new File(toFileName);
+
+        try {
+            newFile.createNewFile();
+        } catch (IOException e) {
+            throw new RuntimeException("Can`t create file", e);
+        }
```
okuzan, line 24: this logic belongs to a method that is responsible for writing data to a file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+        StringBuilder stringBuilder = new StringBuilder()
+                .append("supply").append(SEPARATOR).append(supply).append(System.lineSeparator())
+                .append("buy").append(SEPARATOR).append(buy).append(System.lineSeparator())
+                .append("result").append(SEPARATOR).append(supply - buy);
+
+        return stringBuilder.toString();
+    }
 
+    private void fileWriter(File file, String text) {
```
okuzan, line 69:
```suggestion
    private void writeToFile(String filename, String text) {
```

okuzan (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+                default:
+                    break;
+            }
+        }
+        StringBuilder stringBuilder = new StringBuilder()
+                .append("supply").append(SEPARATOR).append(supply).append(System.lineSeparator())
+                .append("buy").append(SEPARATOR).append(buy).append(System.lineSeparator())
+                .append("result").append(SEPARATOR).append(supply - buy);
+
+        return stringBuilder.toString();
```
olekskov, line 59: it's very rare case when it will add some clarity to introduce new variable and return it immediately.
In most cases you may return result without assignment to variable:
```suggestion
        return new StringBuilder()
                .append("supply").append(SEPARATOR).append(supply).append(System.lineSeparator())
                .append("buy").append(SEPARATOR).append(buy).append(System.lineSeparator())
                .append("result").append(SEPARATOR).append(supply - buy).toSting();
```

olekskov (approved):
Nice, approved, please pay attention to comment below for next tasks

PR #901, opened 2023-03-17
Solution_work-with-file

SerhiUA:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,50 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
okuzan, line 11: image

okuzan (changes requested):
All literals should be constant. GetStatistic should be broken up into smaller methods, which have single responsibility each. In the exception messages, include filenames

PR #905, opened 2023-03-24
completed the task of reading the file, calculating the result and wr…

sshestaka:
…iting to the file

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,77 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
 
+public class WorkWithFile {
+    private int buySum = 0;
+    private int supplySum = 0;
+    private final String supplyOperation = "supply";
+    private final String buyOperation = "buy";
+    private final String result = "result";
+
+    public void getStatistic(String fromFileName, String toFileName) {
+        String[] tempString = readFile(fromFileName).toString().split("\\W+");
+        calculateSum(tempString);
+        checkWriteFile(toFileName);
+        writeFile(toFileName);
+    }
+
+    public StringBuilder readFile(String fromFile) {
+        File file = new File(fromFile);
+        StringBuilder textBox = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFile))) {
+            File checkFile = new File(fromFile);
+            if (checkFile.length() == 0) {
+                System.out.println("File: " + fromFile + "is empty.");
+            }
+            int symbol = 0;
+            while ((symbol = reader.read()) != -1) {
+                textBox.append((char) symbol);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file: " + fromFile, e);
+        }
+        return textBox;
+    }
+
+    public void calculateSum(String[] dataArray) {
+        supplySum = 0;
+        buySum = 0;
+        for (int i = 0; i < dataArray.length; i++) {
+            dataArray[i] = dataArray[i].toLowerCase();
+            if (dataArray[i].equals(supplyOperation)) {
+                supplySum += Integer.parseInt(dataArray[i + 1]);
+            } else if (dataArray[i].equals(buyOperation)) {
+                buySum += Integer.parseInt(dataArray[i + 1]);
+            }
+        }
+    }
+
+    public void checkWriteFile(String toFile) {
+        File checkFile = new File(toFile);
+        if (checkFile.length() != 0) {
+            try (BufferedWriter bufferedWriter =
+                         new BufferedWriter(new FileWriter(toFile, false))) {
+                bufferedWriter.write("");
+            } catch (IOException e) {
+                throw new RuntimeException("Can't write data to file " + toFile, e);
+            }
+        }
+    }
+
+    public void writeFile(String toFile) {
+        File file = new File(toFile);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, false))) {
+            bufferedWriter.write(supplyOperation + "," + supplySum + System.lineSeparator()
+                    + buyOperation + "," + buySum + System.lineSeparator()
+                    + result + "," + (supplySum - buySum) + System.lineSeparator());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file " + toFile, e);
+        }
     }
 }


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+    private final String result = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] tempString = readFile(fromFileName).toString().split("\\W+");
+        calculateSum(tempString);
+        checkWriteFile(toFileName);
+        writeFile(toFileName);
+    }
+
+    public StringBuilder readFile(String fromFile) {
```
andriy-shulepa, line 24: all methods except getStatistic should be private

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+    private int buySum = 0;
+    private int supplySum = 0;
+    private final String supplyOperation = "supply";
+    private final String buyOperation = "buy";
+    private final String result = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] tempString = readFile(fromFileName).toString().split("\\W+");
+        calculateSum(tempString);
+        checkWriteFile(toFileName);
```
andriy-shulepa, line 20: what's the purpose of it? you can just writeToFile without checks

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private int buySum = 0;
+    private int supplySum = 0;
```
andriy-shulepa, line 12:
```suggestion
```
these variables should be local variables

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+
+    public StringBuilder readFile(String fromFile) {
+        File file = new File(fromFile);
+        StringBuilder textBox = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFile))) {
+            File checkFile = new File(fromFile);
+            if (checkFile.length() == 0) {
+                System.out.println("File: " + fromFile + "is empty.");
+            }
+            int symbol = 0;
```
andriy-shulepa, line 32:
```suggestion
            int symbol = 0;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+
 public class WorkWithFile {
+    private int buySum = 0;
+    private int supplySum = 0;
+    private final String supplyOperation = "supply";
+    private final String buyOperation = "buy";
+    private final String result = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] tempString = readFile(fromFileName).toString().split("\\W+");
```
andriy-shulepa, line 18: 1. use more meaningful name for variable
2. create string array like  ["supply,100", "buy,100"...]. with such an approach code will work even if position of array elements was changed

andriy-shulepa (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
sshestaka: OK

src/main/java/core/basesyntax/WorkWithFile.java
sshestaka: I created this check for already existing files...I thought it's a good idea... but then I have changed writeFile method and this check became useless... I left this method because I want to discuss with the mentor this solution for checking if file already exists and isn't empty...

src/main/java/core/basesyntax/WorkWithFile.java
sshestaka: I have defined these variables at the class level because I use them in two methods...OK I can try to return an array from calculateSum method...

src/main/java/core/basesyntax/WorkWithFile.java
sshestaka: I didn't understand what you mean

src/main/java/core/basesyntax/WorkWithFile.java
sshestaka: To be honest, I don't understand what you mean? Could you please explain? Because I received the terms of reference, where the data presentation template was given. And I followed the pattern. If you think that the order of values can change, then there can be a lot of options, for example:
- supply,100
- 100,suplly
- supply,buy,100,200
-...

sshestaka: - "create string array like ["supply,100", "buy,100"...]" - I can do it, but then in the calculateSum method I have to do the same - to split the array, like I already did here:
String[] tempString = readFile(fromFileName).toString().split(System.lineSeparator());

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+    private static final String buyOperation = "buy";
+    private static final String result = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] fromFileData = readFile(fromFileName).toString().split("\\W+");
+        int[] totalAmounts = calculateSum(fromFileData);
+        writeFile(toFileName, totalAmounts);
+    }
+
+    private StringBuilder readFile(String fromFile) {
```
aaananas, line 23:
```suggestion
    private String readFile(String fromFile) {
```

src/main/java/core/basesyntax/WorkWithFile.java
aaananas: remove what is red

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final byte buySum = 0;
+    private static final byte supplySum = 1;
```
aaananas, line 12:
```suggestion
    private static final int BUY_INDEX = 0;
    private static final int SUPPLY_INDEX = 1;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+    private static final String supplyOperation = "supply";
+    private static final String buyOperation = "buy";
+    private static final String result = "result";
```
aaananas, line 15: if you create constants they should be in uppercase
also constants with the same names as the value dont give any better understanding

aaananas (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
sshestaka: Ok

src/main/java/core/basesyntax/WorkWithFile.java
sshestaka: Ok

src/main/java/core/basesyntax/WorkWithFile.java
sshestaka: Ok, I'll try to change it...

Ivan95kos (approved):
For the future, do not write changes in the master branch, create a separate branch for this

sshestaka:
OK! Thank you! I just have forgotten to change the branch...

PR #911, opened 2023-03-31
completed task

sZlishchev:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+    private String[] readFile(String fromFileName) {
+        StringBuilder stringBuilder = new StringBuilder();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(" ");
+                value = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
```
maxline, line 27: please add details about file to the exception message.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+
+    private void writeFile(String[] calculatedStatistic, String toFileName) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            for (String data : calculatedStatistic) {
+                bufferedWriter.write(data);
+                bufferedWriter.write(System.lineSeparator());
+                bufferedWriter.flush();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write file", e);
```
maxline, line 40: please add details about file to the exception message.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+        String[] calculatedStatistic = calculateStatistic(dataFromFile);
+        writeFile(calculatedStatistic, toFileName);
+    }
+
+    private String[] readFile(String fromFileName) {
+        StringBuilder stringBuilder = new StringBuilder();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(" ");
```
maxline, line 23: Space is a magic value here. Introduce a constant (DELIMITER or SEPARATOR) please instead of using hardcoded value. Think about more unique value than space, in general case we can have the sentence in file with words divided by spaces and space doesn't mean that is new line.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write file", e);
+        }
+
+    }
 
+    private String[] calculateStatistic(String[] dataFromFile) {
+        int sumOfSupply = 0;
+        int sumOfBuy = 0;
```
maxline, line 47: variable names can be more concise
```suggestion
        int supplySum = 0;
        int buySum = 0;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+        int ammount = 0;
+        for (String line : dataFromFile) {
+            String[] lineSeparation = line.split(",");
```
maxline, line 50: "," - is magic value, create a constant e.g. CSV_SEPARATOR

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+            ammount = Integer.parseInt(lineSeparation[1]);
```
maxline, line 51: "1" - is magic number, create a constant, e.g. AMOUNT_INDEX, fix the same for "0" magic value

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+        }
+        return stringBuilder.toString().split(" ");
+    }
+
+    private void writeFile(String[] calculatedStatistic, String toFileName) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            for (String data : calculatedStatistic) {
+                bufferedWriter.write(data);
+                bufferedWriter.write(System.lineSeparator());
+                bufferedWriter.flush();
```
maxline, line 37: It is not needed `flush()` data in the loop, it will be very slow down the program execution. 
read about `flush()` e.g. here
https://stackoverflow.com/questions/908168/when-to-flush-a-bufferedwriter
For saving to file optimization, we can create a result concatenated String (remember do not concatenate strings in the loop) and then save the result string to the file just one time.

maxline (changes requested):
Very good job! let's improve the solution

sarakhmen (approved):
Good job :)

PR #915, opened 2023-04-05
My solution

V-Solianyk:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String OPERATION_SUPPLY = "supply";
+
```
maxline, line 11: please remove redundant empty lines between constant names.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+    private static final String COMA_SEPARATOR = ",";
+
+    private static final String TEXT_OF_EXCEPTION = "File not found";
+
+    private static final String REGEX = "\\W";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int supplyAmount = 0;
+        int buyAmount = 0;
+        StringBuilder builder = new StringBuilder();
```
maxline, line 25: Let's try not to use StringBuilder as method parameter and returned result. Better to use String.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+
+    private void writeStringToFile(String toFileName, StringBuilder builder) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(builder.toString());
+        } catch (IOException e) {
+            throw new RuntimeException(TEXT_OF_EXCEPTION);
+        }
+    }
+
+    private void readFromFile(String fromFileName, int supplyAmount, int buyAmount,
```
maxline, line 40: This methdod named as readFromFile but do 3 different work inside
1) read from file
2) calculate statisticts result
3) create the report message 
This three code blocks should be divided to separate methods with appropriate names. One method should do only one logical work.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+                              StringBuilder builder) {
+        String stringWithData;
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            while ((stringWithData = bufferedReader.readLine()) != null) {
+                String[] arrayString = stringWithData.split(REGEX);
+                if (arrayString[0].equals(OPERATION_SUPPLY)) {
+                    supplyAmount += Integer.parseInt(arrayString[1]);
```
maxline, line 47: "1" - is magic number, create a constant, e.g. AMOUNT_INDEX, fix the same for "0" magic value.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+
+    private static final String OPERATION_BUY = "buy";
+
+    private static final String OPERATION_RESULT = "result";
+
+    private static final String COMA_SEPARATOR = ",";
+
+    private static final String TEXT_OF_EXCEPTION = "File not found";
+
+    private static final String REGEX = "\\W";
```
maxline, line 20: let's create more self-explained name, e.g. DELIMITER

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
 public class WorkWithFile {
+    private static final String OPERATION_SUPPLY = "supply";
+
+    private static final String OPERATION_BUY = "buy";
+
+    private static final String OPERATION_RESULT = "result";
+
+    private static final String COMA_SEPARATOR = ",";
+
+    private static final String TEXT_OF_EXCEPTION = "File not found";
```
maxline, line 18: we also need to put more details about the **file** to the exception message. You can use '%' for put value later in the methods, or not using constant.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+                    buyAmount += Integer.parseInt(arrayString[1]);
+                }
+            }
+
+            builder.append(OPERATION_SUPPLY).append(COMA_SEPARATOR).append(supplyAmount)
+                    .append(System.lineSeparator())
+                    .append(OPERATION_BUY).append(COMA_SEPARATOR).append(buyAmount)
+                    .append(System.lineSeparator())
+                    .append(OPERATION_RESULT).append(COMA_SEPARATOR)
+                    .append(supplyAmount - buyAmount);
```
maxline, line 58: 1. This piece of code should be outside of `try catch` block
2. Move this code to the separate method
3. Be attentive about side effect, it is better to put the result of method work into `return` section, instead just of changing the variable which is input parameter.

maxline (changes requested):
Good job! let's improve the solution

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+
+    private String createReportMessage(int supplyAmount, int buyAmount) {
+        return new StringBuilder().append(OPERATION_SUPPLY).append(COMA_SEPARATOR)
+                .append(supplyAmount)
+                .append(System.lineSeparator())
+                .append(OPERATION_BUY).append(COMA_SEPARATOR).append(buyAmount)
+                .append(System.lineSeparator())
+                .append(OPERATION_RESULT).append(COMA_SEPARATOR)
+                .append(supplyAmount - buyAmount)
+                .toString();
```
sarakhmen, line 59: we can use string concatenation here instead of StringBuilder

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+    private void writeStringToFile(String toFileName, String reportMessage) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(reportMessage);
+        } catch (IOException e) {
+            throw new RuntimeException(TEXT_OF_EXCEPTION, e);
+        }
+    }
+
+    private List<String> readFromFile(String fromFileName) {
+        List<String> dataFromBuffer = new ArrayList<>();
```
sarakhmen, line 39: Also let's avoid using collections in this task. We will learn and use them later :)

sarakhmen (approved):
Good job, but you can improve your solution :)

src/main/java/core/basesyntax/WorkWithFile.java
V-Solianyk: but I thought that the builder is generally a better practice when concatenation is needed, we don't create a new object every time, or did I misunderstand something?

PR #916, opened 2023-04-06
Split task by 3 methods

jakop51:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,53 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SPLIT_BY_THIS_STRING = "\\W+";
+    private static final int EMPTY_STRING_LENGTH = 0;
+    private final StringBuilder stringBuilder = new StringBuilder();
+
     public void getStatistic(String fromFileName, String toFileName) {
```
kozhukhovsky, line 14: It is necessary to divide the program logic into 3 methods:
1. Reading data from a file
2. Create a report from this data
3. Writing a report to a file

And you method getStatistic() should only call these three methods

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,57 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SPLIT_BY_THIS_STRING = "\\W+";
+    private static final int EMPTY_STRING_LENGTH = 0;
```
okuzan, line 11: it's self descriptive, I don't think we need it as a field

okuzan (changes requested):
1. Your class has some kind of a shared state - it relies on a StringBuilder field instead of arguments, and methods are void and don't return anything. void methods are notoriously hard to test, and overall this isn't a flexible approach. I suggest your reader method return a String, you then pass that string to createReport, which in turn will return a report which you will pass to the writer method
2. As of now you create one long line and jump through it in a for loop. I think it's a less error-prone and more readable approach to treat lines separately (that means using for example System.lineSeparator() instead of a " " as a custom delimiter). Then you could create a constants for specific positions of operation and amount

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+    private String readInformationFromFile(String fromFileName) {
+        StringBuilder stringBuilder = new StringBuilder();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            String currentStringInFile = bufferedReader.readLine();
+            while (currentStringInFile != null) {
+                stringBuilder.append(currentStringInFile).append(System.lineSeparator());
+                currentStringInFile = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
Ivan95kos, line 25:
```suggestion
            throw new RuntimeException("Can`t read from file: " + fromFileName, e);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+                .append(System.lineSeparator())
+                .append("result,").append(result);
+        return stringBuilder.toString();
+    }
 
+    private void writeInformationToFile(String toFileName, String report) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(report);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
Ivan95kos, line 63:
```suggestion
            throw new RuntimeException("Can`t write to file: " + toFileName, e);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        String fileInfoInString = readInformationFromFile(fromFileName);
+        String report = createReport(fileInfoInString);
+        writeInformationToFile(toFileName, report);
+    }
+
+    private String readInformationFromFile(String fromFileName) {
+        StringBuilder stringBuilder = new StringBuilder();
```
Ivan95kos, line 17: do not use such naming(stringBuilder), how to guess what it stores?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+                    break;
+                case "buy":
+                    buyCounter += Integer.parseInt(buyAndSupplyArray[amountIndex]);
+                    break;
+                default:
+                    break;
+            }
+        }
+        int result = supplyCounter - buyCounter;
+        StringBuilder stringBuilder = new StringBuilder().append("supply,")
```
Ivan95kos, line 50: the same

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+                .append(supplyCounter)
+                .append(System.lineSeparator())
+                .append("buy,").append(buyCounter)
+                .append(System.lineSeparator())
+                .append("result,").append(result);
```
Ivan95kos, line 55: All magic numbers, strings, symbols in your code should be constants.

Ivan95kos (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,68 @@
...
+        }
+        return fileContent.toString();
+    }
+
+    private String createReport(String fileInString) {
+        final int operationTypeIndex = 0;
+        final int amountIndex = 1;
+        final String supplyString = "supply,";
+        final String buyString = "buy,";
+        final String resultString = "result,";
```
okuzan, line 35: these should be class level constants, in CONSTANT_CASE

okuzan (changes requested):
Good job, constant extraction left

PR #917, opened 2023-04-06
jv work with file solution

furthernull:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int OPERATION = 0;
+    private static final int AMOUNT = 1;
+    private static final String DATA_SEPARATOR = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
```
kozhukhovsky, line 14: It is necessary to divide the program logic into 3 methods:
1. Reading data from a file
2. Create a report from this data
3. Writing a report to a file

And you method getStatistic() should only call these three methods

kozhukhovsky (changes requested):

olekskov (approved):
Nice job! 
I'd use if/else instead of switch for binary choice but it's matter of taste

PR #922, opened 2023-04-10
Solution

Stipaxa-hub:
Created readData method. This method will read all data in csv file and return string with all data in file.

Created createReport method. This method will create a report and return string report from market after working day. Implemented method getStatistic. This method use readData and createReport methods. Also, this method will write report, which we will get from createReport method, to new file.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,56 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
 
+public class WorkWithFile {
+    private static final int POSITION_OPERATION_TYPE = 0;
+    private static final int POSITION_AMOUNT = 1;
+
+    public void getStatistic(String fromFileName, String toFileName) {
+        File file = new File(toFileName);
+        String report = createReport(fromFileName);
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(file))) {
+            writer.write(report);
+            writer.flush();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file ", e);
+        }
+    }
+
+    private String createReport(String fromFileName) {
+        int supply = 0;
+        int buy = 0;
+        String[] lines = readData(fromFileName).split(System.lineSeparator());
+        for (String line : lines) {
+            String[] data = line.split(",");
+            if (data[POSITION_OPERATION_TYPE].equals("supply")) {
+                supply += Integer.parseInt(data[POSITION_AMOUNT]);
+            }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+    private static final int POSITION_AMOUNT = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        File file = new File(toFileName);
+        String report = createReport(fromFileName);
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(file))) {
+            writer.write(report);
+            writer.flush();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file ", e);
```
okuzan, line 21: include filename to message

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+        }
+    }
+
+    private String createReport(String fromFileName) {
+        int supply = 0;
+        int buy = 0;
+        String[] lines = readData(fromFileName).split(System.lineSeparator());
+        for (String line : lines) {
+            String[] data = line.split(",");
```
okuzan, line 30: separator constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+        File file = new File(fromFileName);
+        StringBuilder builder = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(file))) {
+            String value = reader.readLine();
+            while (value != null) {
+                builder.append(value).append(System.lineSeparator());
+                value = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file ", e);
```
okuzan, line 52: include filename

okuzan (changes requested):
Divide your logic further. create report shouldn't deal with reading files, it should be given only content. Extract logic dealing with file writing to a separate method too. Thus, your main public method will be composed of just three method calls

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        writeData(createReport(fromFileName), toFileName);
+    }
+
+    private String createReport(String fromFileName) {
+        int supply = 0;
+        int buy = 0;
+        String[] lines = readData(fromFileName).split(System.lineSeparator());
+        for (String line : lines) {
+            String[] data = line.split(SEPARATOR);
```
olekskov, line 26:
```suggestion
    public void getStatistic(String fromFileName, String toFileName) {
        writeData(createReport(readData(fromFileName)), toFileName);
    }

    private String createReport(String fileContent) {
        int supply = 0;
        int buy = 0;
        String[] lines = fileContent.split(System.lineSeparator());
        for (String line : lines) {
            String[] data = line.split(SEPARATOR);
```

olekskov (approved):
Good job, approved.
Please take into account comment below. Better isolation of methods increases reusability in future

PR #927, opened 2023-04-22
I decided to use try-with-resources to be unsure that the BufferedRea…

RostyslavOnysh:
…der and BufferedWriter object are closed after use.

for reading file I use BufferedReader
for write data to file I use BufferedWriter.

Created variables in which I save Supply value and Buy value.

Splits each line into two parts separated by comma.First part save as string variable "operation" and second one as a integer and store in "amount" variable.

I use switch statment to increment totalSupply or totalBuy.

After int variable "result" computes difference between totalSupply and totalBuy.

In the end using write() method to write info to new file.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,38 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
-
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
+                BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName))) {
+            int totalSupply = 0;
+            int totalBuy = 0;
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] parts = line.split(",");
+                String operation = parts[0].trim();
+                int amount = Integer.parseInt(parts[1].trim());
+                switch (operation) {
+                    case "supply":
+                        totalSupply += amount;
+                        break;
+                    case "buy":
+                        totalBuy += amount;
+                        break;
+                    default:
+                }
+            }
+            int result = totalSupply - totalBuy;
+            writer.write("supply," + totalSupply + System.lineSeparator());
+            writer.write("buy," + totalBuy + System.lineSeparator());
+            writer.write("result," + result);
+        } catch (IOException e) {
+            throw new RuntimeException("Error reading/writing file", e);
+        }
     }
 }


RostyslavOnysh:
I tried to complete this task using a for loop, but I kept getting errors. Can you help me understand how to implement it correctly?  And does it make any sense to use a for loop instead of a switch statement?
I use this block of code : 
if (operation.equals("supply")) {
                totalSupply += amount;
            } else if (operation.equals("buy")) {
                totalBuy += amount;
            }

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,38 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
Ivan95kos, line 10: this method should contain three calls of other methods: readFile() - returns content, generateReport() -(you pass contents here), writeToFile()-(you pass report and filename here ), this will be a good code decomposition

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,38 @@
...
+                String[] parts = line.split(",");
+                String operation = parts[0].trim();
+                int amount = Integer.parseInt(parts[1].trim());
+                switch (operation) {
+                    case "supply":
+                        totalSupply += amount;
+                        break;
+                    case "buy":
+                        totalBuy += amount;
+                        break;
```
Ivan95kos, line 26: All magic numbers, strings, symbols in your code should be constants.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,38 @@
...
+                    default:
+                }
+            }
+            int result = totalSupply - totalBuy;
+            writer.write("supply," + totalSupply + System.lineSeparator());
+            writer.write("buy," + totalBuy + System.lineSeparator());
+            writer.write("result," + result);
```
Ivan95kos, line 33: the same, magic strings

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,38 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Error reading/writing file", e);
```
Ivan95kos, line 35:
```suggestion
            throw new RuntimeException("Can't not read file:" + fromFileName, e);
```
and should be
```suggestion
            throw new RuntimeException("Can't not write data to file:" + toFileName, e);
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+    }
+
+    private String generateReport(String contents) {
+        int totalSupply = 0;
+        int totalBuy = 0;
+        String[] lines = contents.trim().split(System.lineSeparator());
+        for (String line : lines) {
+            String[] splitter = line.split(SEPARATE_COMA);
+            String operation = splitter[0].trim();
+            int amount = Integer.parseInt(splitter[1].trim());
```
okuzan, line 41: these two should be constants -  OPERATION_INDEX, AMOUNT_INDEX

okuzan (changes requested):
Good job!

PR #937, opened 2023-04-27
Added task realization

pkobevko:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,88 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
 public class WorkWithFile {
+    private static final int INDEX_OF_SUPPLY_AMOUNT = 0;
+    private static final int INDEX_OF_BUY_AMOUNT = 1;
+    private static final int INDEX_OF_OPERATION_TYPE = 0;
+    private static final int INDEX_OF_AMOUNT = 1;
```
rovein, line 17: Looks like duplicated constants, can we keep only two of them?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,88 @@
...
+    private static final String COMMA_SEPARATOR = ",";
+    private static final String SUPPLY_TYPE_STRING = "supply";
+    private static final String BUY_TYPE_STRING = "buy";
+    private static final String RESULT_REPORT_STRING = "result";
```
rovein, line 21: Don't use suffix `STRING`. Also, `TYPE` can be replaced to `OPERATION`.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,88 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int[] data = readDataFromFile(fromFileName);
+        String report = createReportFromData(data);
+        writeReportIntoFile(report, toFileName);
```
rovein, line 26: Good, but let's reorder the method's declaration in the code according to the order of their calls. It should be easy to read the code like a book from up to down.

Reorder so:
- readDataFromFile
- parseDataLine
- createReportFromData
- writeReportIntoFile
- deleteIfExists

Then take a look and think if it improved the readability. See Clean Code of Robert Martin for more details.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,88 @@
...
+    }
+
+    private void writeReportIntoFile(String report, String toFileName) {
+        File file = new File(toFileName);
```
rovein, line 30: You don't need it.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,88 @@
...
+        deleteFileIfExists(file, toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, true))) {
+            bufferedWriter.write(report);
+        } catch (IOException e) {
+            throw new RuntimeException("Error writing report into file", e);
+        }
+    }
+
+    private void deleteFileIfExists(File file, String toFileName) {
```
rovein, line 39: Unused param `file`, remove it.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,88 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        int[] data = readDataFromFile(fromFileName);
+        String report = createReportFromData(data);
+        writeReportIntoFile(report, toFileName);
+    }
+
+    private void writeReportIntoFile(String report, String toFileName) {
+        File file = new File(toFileName);
+        deleteFileIfExists(file, toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file, true))) {
```
rovein, line 32:
```suggestion
        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName, true))) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,88 @@
...
+        StringBuilder stringBuilder = new StringBuilder();
+        int supplyAmount = data[INDEX_OF_SUPPLY_AMOUNT];
+        int buyAmount = data[INDEX_OF_BUY_AMOUNT];
+        int result = supplyAmount - buyAmount;
+        stringBuilder.append(SUPPLY_TYPE_STRING).append(COMMA_SEPARATOR)
+                .append(supplyAmount).append(System.lineSeparator());
+        stringBuilder.append(BUY_TYPE_STRING).append(COMMA_SEPARATOR)
+                .append(buyAmount).append(System.lineSeparator());
+        stringBuilder.append(RESULT_REPORT_STRING).append(COMMA_SEPARATOR)
+                .append(result);
```
rovein, line 57: Let's chain all `append` together, without additional `stringBuilder.append`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,88 @@
...
+            String line = bufferedReader.readLine();
+            while (line != null) {
+                parseDataLine(data, line);
+                line = bufferedReader.readLine();
+            }
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("File not found", e);
+        } catch (IOException e) {
+            throw new RuntimeException("Error reading data from file", e);
+        }
```
rovein, line 74: Don't create two identical `catch` blocks. Either use `multiCatch` ([reference](https://jenkov.com/tutorials/java-exception-handling/catching-multiple-exceptions.html)).

But, here you have a hierarchy. `FileNotFound` is already an `IOException`. So remove the first `catch`.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,88 @@
...
+                .append(result);
+        return stringBuilder.toString();
+    }
+
+    private int[] readDataFromFile(String fromFileName) {
+        int[] data = new int[2];
+        File file = new File(fromFileName);
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
+            String line = bufferedReader.readLine();
+            while (line != null) {
```
rovein, line 66:
```suggestion
            String line;
            while ((line = reader.readLine()) != null) {
```

rovein (changes requested):

rovein (approved):
Good job! Keep the rhythm!

PR #940, opened 2023-04-27
implemented getStatistic(String fromFileName, String toFileName) method

ViktoriiaSemenchenko:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+    private static final String SUPPLY_DATA = "supply";
+    private static final String BUY_DATA = "buy";
+    private static final String RESULT_DATA = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        writeDataToFile(toFileName, getReport(readFromFile(fromFileName)));
+    }
+
+    private String readFromFile(String fileName) {
+        File file = new File(fileName);
```
rovein, line 24: Let's remove it, see next comment.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+        StringBuilder builder = new StringBuilder();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
```
rovein, line 26:
```suggestion
        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+            String dataFromFile = bufferedReader.readLine();
+            while (dataFromFile != null) {
+                builder.append(dataFromFile).append(System.lineSeparator());
+                dataFromFile = bufferedReader.readLine();
+            }
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("File not found" + fileName, e);
+        } catch (IOException e) {
+            throw new RuntimeException("Can not read file" + fileName, e);
+        }
```
rovein, line 36: Don't create two identical `catch` blocks. Either use `multiCatch` ([reference](https://jenkov.com/tutorials/java-exception-handling/catching-multiple-exceptions.html)).

But, here you have a hierarchy. `FileNotFound` is already an `IOException`. So remove the first `catch`.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+                .append(supply)
+                .append(System.lineSeparator())
+                .append(BUY_DATA)
+                .append(SEPARATOR_OF_DATA)
+                .append(buy)
+                .append(System.lineSeparator())
+                .append(RESULT_DATA)
+                .append(SEPARATOR_OF_DATA)
+                .append(result)
+                .toString();
```
rovein, line 66: Let's extract it to the separate `private` method. Remember from `checklist`:

**Divide program logic.** For example: “write to file” or “create report” - are a separate operations. So we should create separate methods for them.
```suggestion
        return createResultString(...);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+    }
 
+    private void writeDataToFile(String toFileName, String report) {
+        File file = new File(toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file))) {
```
rovein, line 71: The same.
```suggestion
        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
```

rovein (changes requested):

PR #943, opened 2023-04-27
add getStatistic method solution

AnastasiiaVariuchuk:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+        int buyAllSum = 0;
+        for (int i = 0; i < stringsFromFile.length - 1; i += 2) {
+            if (stringsFromFile[i].equals(BUY_KEYWORD)) {
+                buyAllSum += Integer.parseInt(stringsFromFile[i + 1]);
+            } else {
+                supplyAllSum += Integer.parseInt(stringsFromFile[i + 1]);
+            }
+        }
+        String[] resultStrings = new String[]{"supply,"
+                + supplyAllSum, "buy," + buyAllSum, "result," + (supplyAllSum - buyAllSum)};
```
rovein, line 35: Let's extract it to the separate `private` method. Remember from `checklist`:

**Divide program logic.** For example: “write to file” or “create report” - are a separate operations. So we should create separate methods for them.

Also, let's build a single result String, not the array.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            for (String str : resultStrings) {
+                bufferedWriter.write(str);
+                bufferedWriter.write(System.lineSeparator());
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write in the file: ", e);
+        }
```
rovein, line 43: The same, let's extract all this `try/catch` to the separate `private` method.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+        String[] stringsFromFile = stringBuilder.toString().split(COMMA_SEPARATOR);
+        int supplyAllSum = 0;
+        int buyAllSum = 0;
+        for (int i = 0; i < stringsFromFile.length - 1; i += 2) {
+            if (stringsFromFile[i].equals(BUY_KEYWORD)) {
+                buyAllSum += Integer.parseInt(stringsFromFile[i + 1]);
+            } else {
+                supplyAllSum += Integer.parseInt(stringsFromFile[i + 1]);
+            }
+        }
```
rovein, line 33: So, as I see, first you are reading all lines and saving them to `StringBuilder` (lines 14 - 23), then you are processing them again and calculating the `supply` and `buy` sum. Could we make it easier?

Can we calculate the sum on the lines 14-23, without saving to StringBuilder?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+        String[] resultStrings = new String[]{"supply,"
+                + supplyAllSum, "buy," + buyAllSum, "result," + (supplyAllSum - buyAllSum)};
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            for (String str : resultStrings) {
+                bufferedWriter.write(str);
+                bufferedWriter.write(System.lineSeparator());
```
rovein, line 39: Is the last line separator really needed here?

rovein (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,43 @@
...
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String BUY_KEYWORD = "buy";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        writeDataToFile(calculateReadDataFromFile(fromFileName), toFileName);
+    }
+
+    private String calculateReadDataFromFile(String fromFileName) {
```
sarakhmen, line 16: This method has an unclear purpose, let's divide it into two separate: first one will read data from file and second one will process that data

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,43 @@
...
+        int supplyAllSum = 0;
+        int buyAllSum = 0;
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            String str = bufferedReader.readLine();
+            while (str != null) {
+                if (str.contains(BUY_KEYWORD)) {
+                    buyAllSum += Integer.parseInt(str.replaceAll("[a-zA-Z.,_-]", ""));
```
sarakhmen, line 23: consider using split() method instead

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,43 @@
...
+                } else {
+                    supplyAllSum += Integer.parseInt(str.replaceAll("[a-zA-Z.,_-]", ""));
+                }
+                str = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file: ", e);
```
sarakhmen, line 30: you should also add information about filename to your exception message

sarakhmen (changes requested):
Let's improve your solution ;)

PR #963, opened 2023-05-08
Counted all supply & buy, counted them in result and write it in file.

Monsterhunter525:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,42 @@
...
 
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
Rommelua, line 11: `getStatistic` should have 3 lines - 3 private method calls one by one
image

Rommelua (changes requested):
https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file

Before you create PR with your homework, you need to go through the checklist under the task and correct all the points described there. The mentor will not check the work until the checklist points are corrected.
Do not hesitate to contact us in Slack if you do not understand something)

Rommelua (changes requested):
Do not use static methods here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,52 @@
...
+        writeToFile(toFileName, report);
+    }
+
+    private String createReport(List<String> dataFromFile) {
+        int supplies = 0;
+        int buys = 0;
+        int result;
+        for (String content : dataFromFile) {
+            String action = content.split(",")[0];
+            String sum = content.split(",")[1];
```
andriy-shulepa, line 23: extract indexes to constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,52 @@
...
+            if (action.equals("supply")) {
+                supplies += Integer.parseInt(sum);
+            }
+            if (action.equals("buy")) {
```
andriy-shulepa, line 27: extract actions to constants. and it's better to check like "supply".equals(action) since action can be null and you may catch NullPointerException with first approach

andriy-shulepa (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
Monsterhunter525: Thanks! Thats a good one.

andriy-shulepa (approved):
next time, create new branch and push your solution there instead of pushing to the master branch

PR #965, opened 2023-05-10
The getStatistic method takes two strings: fromFileName and toFileNam…

kaptetos:
…e, which are the names of the input and output files, respectively. The method throws an IOException, which means that I/O errors may occur when working with files.

The method creates two variables supplyTotal and buyTotal, which will be used to calculate the sum of the values of the amount field corresponding to the supply or buy operation. Then two try-with-resources blocks are created, in each of which the corresponding objects for reading and writing files are created.

Using the BufferedReader object and the readLine() method, lines are read sequentially from the input file. Each line is split into fields using the split() method, and then the number of fields is checked to avoid errors if the input data is incorrectly formatted. If the number of fields is 2, then a check is made which operation is specified in the first field, and the amount value is added to the corresponding variable.

Next, the value of the result variable is calculated by subtracting the value of buyTotal from supplyTotal.

Then, using the BufferedWriter object and the write() method, three lines are written to the output file containing the analysis results in the format described in the problem statement.

Thus, this code reads data from the input file, counts the number of supply and buy operations, and then writes the analysis report to the output file.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,38 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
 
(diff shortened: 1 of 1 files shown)


okuzan (changes requested):
make sure your build passes before requesting a review 😉

PR #969, opened 2023-05-13
Hometask WorkWithFile completed

MarynaKushniruk:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final String WHITESPACE = " ";
+    private static final String COMA = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
```
Ivan95kos, line 16: this method should contain three calls of other methods: readFile() - returns content, generateReport() -(you pass contents here), writeToFile()-(you pass report and filename here ), this will be a good code decomposition

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+        StringBuilder builder = new StringBuilder();
```
Ivan95kos, line 17: do not use such naming, how to guess what it stores?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName));
+            String words = bufferedReader.readLine();
+            while (words != null) {
+                builder.append(words).append(" ");
+                words = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file", e);
```
Ivan95kos, line 26:
```suggestion
            throw new RuntimeException("Can't read the file: " + fromFileName, e);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
 
+        try {
+            BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName));
+            for (String data : reportFinal) {
+                bufferedWriter.write(data);
+                bufferedWriter.newLine();
+                bufferedWriter.flush();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file", e);
```
Ivan95kos, line 53:
```suggestion
            throw new RuntimeException("Can't write data to file: " + toFileName, e);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+                sumSupply += Integer.parseInt(reportData[1]);
+            } else if (reportData[0].equals(BUY)) {
+                sumBuy += Integer.parseInt(reportData[1]);
+            }
+        }
+        int result = sumSupply - sumBuy;
+        String[] reportFinal = new String[3];
+        reportFinal[0] = SUPPLY + COMA + sumSupply;
+        reportFinal[1] = BUY + COMA + sumBuy;
+        reportFinal[2] = RESULT + COMA + result;
```
Ivan95kos, line 43: It is better to use constants instead of numbers

Ivan95kos (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file: " + fromFileName, e);
+        }
+        return readedData.toString().split(WHITESPACE);
+    }
+
+    public String[] generateReport(String[] readedData) {
+        int sumSupply = ZERO;
+        int sumBuy = ZERO;
```
olekskov, line 42: 0 is default value for int, no need to explicitly set it
```suggestion
        int sumSupply;
        int sumBuy;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final String WHITESPACE = " ";
+    private static final String COMA = ",";
+    private static final int ZERO = 0;
```
olekskov, line 15: Please use appropriate naming for constants that represents meaning of value but not value itself:
```suggestion
    private static final int OPERATION_TYPE_INDEX = 0;
```
Please recheck all usages of ZERO, ONE, TWO. You may require different constants for same value if this value is used for different purposes

olekskov (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
MarynaKushniruk: I left ZERO initialization because code gives compilation mistake than..

src/main/java/core/basesyntax/WorkWithFile.java
MarynaKushniruk: Change

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+        String[] reportFinal = new String[3];
+        reportFinal[INDEX_ZERO] = SUPPLY + COMA + sumSupply;
+        reportFinal[INDEX_ONE] = BUY + COMA + sumBuy;
+        reportFinal[INDEX_TWO] = RESULT + COMA + result;
+        return reportFinal;
+    }
 
+    public void writeToFile(String[] readyReport, String toFileName) {
+        try {
+            BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName));
```
okuzan, line 62: Use try-with-resources to automatically close BufferedWriter.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+    private static final int INDEX_TWO = 2;
+    private static final int ZERO = 0;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] readedData = readFile(fromFileName);
+        String[] readyReport = generateReport(readedData);
+        writeToFile(readyReport, toFileName);
+    }
+
+    public String[] readFile(String fromFileName) {
```
okuzan, line 26: Consider changing method visibility to private, as it's not used outside of the class.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+        StringBuilder readedData = new StringBuilder();
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName));
```
okuzan, line 29: Use try-with-resources to automatically close BufferedReader.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+        }
+        int result = sumSupply - sumBuy;
+        String[] reportFinal = new String[3];
+        reportFinal[INDEX_ZERO] = SUPPLY + COMA + sumSupply;
+        reportFinal[INDEX_ONE] = BUY + COMA + sumBuy;
+        reportFinal[INDEX_TWO] = RESULT + COMA + result;
+        return reportFinal;
+    }
 
+    public void writeToFile(String[] readyReport, String toFileName) {
```
okuzan, line 60: Consider changing method visibility to private, as it's not used outside of the class.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+                readedData.append(words).append(" ");
+                words = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file: " + fromFileName, e);
+        }
+        return readedData.toString().split(WHITESPACE);
+    }
+
+    public String[] generateReport(String[] readedData) {
```
okuzan, line 41: Consider changing method visibility to private, as it's not used outside of the class.

src/main/java/core/basesyntax/WorkWithFile.java
okuzan: not fixed

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+                sumSupply += Integer.parseInt(reportData[INDEX_ONE]);
+            } else if (reportData[INDEX_ZERO].equals(BUY)) {
+                sumBuy += Integer.parseInt(reportData[INDEX_ONE]);
+            }
+        }
+        int result = sumSupply - sumBuy;
+        String[] reportFinal = new String[3];
+        reportFinal[INDEX_ZERO] = SUPPLY + COMA + sumSupply;
+        reportFinal[INDEX_ONE] = BUY + COMA + sumBuy;
+        reportFinal[INDEX_TWO] = RESULT + COMA + result;
```
okuzan, line 56: why to use array here? Just create a one string report and return it, then write it with bufferedWriter.write() and that's it :)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file: " + fromFileName, e);
+        }
+        return readedData.toString().split(WHITESPACE);
+    }
+
+    public String[] generateReport(String[] readedData) {
+        int sumSupply = ZERO;
+        int sumBuy = ZERO;
```
okuzan, line 43:
```suggestion
        int sumSupply = 0;
        int sumBuy = 0;
```
it doesn't hold any special meaning, it's okay to hardcode 0 to initialize a variable

okuzan (changes requested):

PR #970, opened 2023-05-14
Was gotten stream of String by Files.lines method

pochtalon:
Was created method getSring() for getting result string from stream
   Were created String constant inside
   Were gotten supply and bue values
   Was Gotten result String by StringBuilder
Was written result String into the file by Files.write() method

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,44 @@
 package core.basesyntax;
 
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        Stream<String> str = null;
+        try {
+            str = Files.lines(Path.of(fromFileName));
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        String toWrite = getString(str);
+        try {
+            Files.write(Path.of(toFileName), toWrite.getBytes(StandardCharsets.UTF_8));
+        } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Your trying to write bytes to file was not successful", e);
+        }
+    }
 
+    private String getString(Stream<String> stream) {
+        final String Supply = "supply";
+        final String Buy = "buy";
+        final String Result = "result";
+        final String Delimiter = ",";
```
okuzan, line 31: make these class constants , not just final local variables

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+        Map<String, Integer> csvSum = stream
+                .map(e -> e.split(","))
+                .collect(Collectors.groupingBy(e -> e[0],
+                        Collectors.summingInt(e -> Integer.valueOf(e[1]))));
```
okuzan, line 35: indexes should be constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+        }
+        String toWrite = getString(str);
+        try {
+            Files.write(Path.of(toFileName), toWrite.getBytes(StandardCharsets.UTF_8));
+        } catch (IOException e) {
+            throw new RuntimeException("Your trying to write bytes to file was not successful", e);
+        }
+    }
 
+    private String getString(Stream<String> stream) {
```
okuzan, line 27: think of a more descriptive name that resembles the purpose of this method

okuzan (changes requested):
Let's restructure your code
Designate one method for reading from file, one for processing the data, one for writing a report
image

olekskov (approved):
Good job

PR #973, opened 2023-05-17
created logic for getStatistics method

DZarudnyi:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+            String[] readValue;
+
+            while (readLine != null) {
+                readValue = readLine.split(",");
+                if (readValue[0].equals(STATISTICS_FIELDS[SUPPLY_POS])) {
+                    statisticsValues[0] += Integer.parseInt(readValue[1]);
+                } else {
+                    statisticsValues[1] += Integer.parseInt(readValue[1]);
+                }
+                readLine = bufferedReader.readLine();
```
kozhukhovsky, line 34: The read-from-file method should simply read the data from the file and return it. The logic of creating a report must be moved to a separate method.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String[] STATISTICS_FIELDS = {"supply", "buy", "result"};
+    private static final int SUPPLY_POS = 0;
```
kozhukhovsky, line 11:
```suggestion
    private static final int SUPPLY_POSITION = 0;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+    private static final int BUY_POS = 1;
```
kozhukhovsky, line 12:
```suggestion
    private static final int BUY_POSITION = 1;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+    private static final int RESULT_POS = 2;
```
kozhukhovsky, line 13:
```suggestion
    private static final int RESULT_POSITION = 2;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+                readValue = readLine.split(",");
+                if (readValue[0].equals(STATISTICS_FIELDS[SUPPLY_POS])) {
+                    statisticsValues[0] += Integer.parseInt(readValue[1]);
+                } else {
+                    statisticsValues[1] += Integer.parseInt(readValue[1]);
+                }
+                readLine = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            System.out.println("Error while reading a file");
```
kozhukhovsky, line 37: It is better to throw the exception further, and so we displayed it on the screen, and then some logic is done and the result is returned, but it should not

kozhukhovsky (changes requested):

DZarudnyi:
Okay, so i divided everything in separate methods, even tho this solution has some of its own flaws, i can see why you told me to do this.
Another question that i have rn, is my main method getStatistics, where i splitted everything in a way, you would want me to, e.g.:
     String fileContent = readFromFile(fromFileName);
      int[] statisticsValues = calculateStatistics(fileContent);
      writeToFile(statisticsValues, toFileName);
When i could`ve just:
      writeToFile(calculateStatistics(readFromFile(fromFileName)), toFileName);
If to use this approach, we won`t allocate unnecessary memory and it`s still readable. So is this a good approach or not?

src/main/java/core/basesyntax/WorkWithFile.java
olekskov: not fixed

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+    private static final int RESULT_POSITION = 2;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String fileContent = readFromFile(fromFileName);
+        int[] statisticsValues = calculateStatistics(fileContent);
+        writeToFile(statisticsValues, toFileName);
+    }
+
+    private String readFromFile(String fromFileName) {
+        StringBuilder sb = new StringBuilder();
```
olekskov, line 23: Please avoid using 1-2 letters variable names. And names that based on class. Name should describe what data is inside:
```suggestion
        StringBuilder fileContent = new StringBuilder();
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+
+        return sb.toString().trim();
+    }
+
+    private int[] calculateStatistics(String fileContent) {
+        int[] statisticsValues = new int[3];
+
+        Scanner scanner = new Scanner(fileContent);
+        while (scanner.hasNextLine()) {
+            String[] line = scanner.nextLine().split(",");
```
olekskov, line 44: It seems Scanner doesn't provide any benefits in our case compared to simple for/each cycle
```suggestion
        for (String line : fileContent.split(System.lineSeparator()) {
            String[] parsedRecord = line().split(",");
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+            if (line[0].equals(STATISTICS_FIELDS[SUPPLY_POSITION])) {
+                statisticsValues[SUPPLY_POSITION] += Integer.parseInt(line[1]);
+            } else {
+                statisticsValues[BUY_POSITION] += Integer.parseInt(line[1]);
```
olekskov, line 48: These hardcoded 0/1 indexes also will look better when extracted to constants with proper naming.
N.B.
Please do not reuse existing constant for another purpose even if it has same value

olekskov (changes requested):

olekskov:
> Okay, so i divided everything in separate methods, even tho this solution has some of its own flaws, i can see why you told me to do this.
> Another question that i have rn, is my main method getStatistics, where i splitted everything in a way, you would want me to, e.g.:
> `String fileContent = readFromFile(fromFileName); int[] statisticsValues = calculateStatistics(fileContent); writeToFile(statisticsValues, toFileName);`
> When i could've just:
> `writeToFile(calculateStatistics(readFromFile(fromFileName)), toFileName);`
> If to use this approach, we won`t allocate unnecessary memory and it's still readable. So is this a good approach or not?

Memory allocation difference between these two approaches is negligible. Main factor to decide what would be better is what would be more readable. There is no 100% right answer here what is better.
If you fill that construction
**writeToFile(calculateStatistics(readFromFile(fromFileName)), toFileName);**
is ok to be clearly understood by another person and it's shorter then it's ok.
If you have doubts then it's better to split for more simple operations with possibility to introduce variable with proper naming for even more clarity

src/main/java/core/basesyntax/WorkWithFile.java
DZarudnyi: I am sorry, but creating own constant for every number in code just seems absurd. I see that you mean better code readability by this, but simple comment, in that exact case, would explain everything much better, in my opinion.

kozhukhovsky: comments in the code are a very bad practice, if there is a need to leave a comment, then something is wrong with the code, well, or some very tricky construction that cannot be done in any other way, then it is acceptable.
Usually the code itself should be readable. But such numbers, I have no idea what is there at index 1. Therefore, it is great to give names so that it is clear what exactly we will get

PR #977, opened 2023-05-23
add my solution

mykhailoKrlKrk:


okuzan (changes requested):
divide your logic into several parts. You should deal with reading file, processing data, writing results to a file separately. public method getStatistic itself should be ~ 3 lines - calling other utility methods. All literals should be constant - strings, indexes. Exception messages should be informative - include filenames

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        int supply = 0;
+        int buy = 0;
+
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            String currentLine;
+            while ((currentLine = bufferedReader.readLine()) != null) {
+                String[] block = currentLine.split(SEPARATED_CHARACTER);
+                String operationType = block[0].trim();
+                int amount = Integer.parseInt(block[1].trim());
```
okuzan, line 24: these should be OPERATION_INDEX, AMOUNT_INDEX

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
+    private static final String SEPARATED_CHARACTER = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int supply = 0;
+        int buy = 0;
+
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            String currentLine;
+            while ((currentLine = bufferedReader.readLine()) != null) {
+                String[] block = currentLine.split(SEPARATED_CHARACTER);
```
okuzan, line 22: reading should be done in a dedicated private method, not here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final String SEPARATED_CHARACTER = ",";
```
okuzan, line 13:
```suggestion
    private static final String SEPARATOR = ",";
```

okuzan: and it's not a char  :)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
+            }
+
+            int result = supply - buy;
+            writeStatisticsToFile(toFileName, supply, buy, result);
+        } catch (IOException e) {
+            throw new RuntimeException("Error with reading file: " + fromFileName, e);
+        }
+    }
 
+    private void writeStatisticsToFile(String toFileName, int supply, int buy, int result) {
```
okuzan, line 40: this methods does two things - generate a report and writes it. Separate concerns - method should have a ready report string for an input

okuzan (changes requested):

olekskov (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final String SEPARATOR = ",";
+    private static final int OPERATION_INDEX = 0;
+    private static final int AMOUNT_INDEX = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] textFromFile = readStatisticFromFile(fromFileName);
+
```
Ivan95kos, line 21:
```suggestion
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
```
Ivan95kos, line 14: The name of the constant should not be the same as the value being stored

Ivan95kos (approved):
Great job, with minor comments

PR #979, opened 2023-05-26
no comment yet

Olena-Pro:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int INDEX_OPERATION = 0;
+    private static final int INDEX_DATA = 1;
+    private static final String LIMIT_FIELD = ",";
+    private String reportText;
+
     public void getStatistic(String fromFileName, String toFileName) {
```
Ivan95kos, line 16: this method should contain three calls of other methods: readFile() - returns content, generateReport() -(you pass contents here), writeToFile()-(you pass report and filename here ), this will be a good code decomposition

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+        String nameSupply = "supply";
+        String nameBuy = "buy";
```
Ivan95kos, line 18: use constants instead of strings

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+        StringBuilder stringBuilder = new StringBuilder();
```
Ivan95kos, line 19: do not use such naming, how to guess what it stores?

Ivan95kos (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,80 @@
...
+    private static final int INDEX_DATA = 1;
+    private static final String LIMIT_FIELD = ",";
+    private static final String OPEERATION_ARRIVAL = "supply";
+    private static final String OPERATION_EXPENSE = "buy";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String inputInfo;
+        String reportText;
+        inputInfo = readFile(fromFileName);
+        reportText = generateReport(inputInfo);
```
okuzan, line 21:
```suggestion
        String inputInfo = readFile(fromFileName);
        String reportText = generateReport(inputInfo);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,80 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file " + fromFileName, e);
+        }
+        return stringBuilder.toString();
+    }
+
+    private String generateReport(String inputInfo) {
+        if (inputInfo.length() == 0) {
+            return null;
+        }
```
okuzan, line 42: returning null silently is a bad practice, better throw an exception if we don't expect this situation. In this task, we don't need to validate input

okuzan (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,80 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int INDEX_OPERATION = 0;
+    private static final int INDEX_DATA = 1;
```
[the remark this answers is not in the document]
okuzan: this is a minor change, but I would recommend it nevertheless :)

okuzan (approved):
Good job

PR #984, opened 2023-06-04
implemented getStatistic method that read text from file and records …

happytroya:
…it to String, then splitted it by lines, then by name and value. This method then writes info to file in csv format

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,63 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
 public class WorkWithFile {
+    private static final String SPLITTER = ",";
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final int NAME_INDEX = 0;
+    private static final int VALUE_INDEX = 1;
+    private StringBuilder textToWrite = new StringBuilder();
+
     public void getStatistic(String fromFileName, String toFileName) {
```
olekskov, line 18: It isn't good practice to make long methods, especially public ones.
Please split your solution to three private methods like readFile, generateReport and writeToFile.
Then in your public method make only calls to there 3 methods to make it short and clear

olekskov (changes requested):
Good job overall, please enhance solution according to comments

PR #986, opened 2023-06-06
I would like to get feedback about my mistakes, because the tests pass

NSashchenko:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,55 @@
...
+    private static final String OPERATION_SUPPLY = "supply";
+    private static final String OPERATION_BUY = "buy";
+    private static final String REPORT_HEADER_SUPPLY = "supply";
+    private static final String REPORT_HEADER_BUY = "buy";
+    private static final String REPORT_HEADER_RESULT = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
+            BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName));
```
olekskov, line 20: Please use try-with-resources for autoclosable resources

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,55 @@
...
+
 public class WorkWithFile {
+    private static final String DELIMITER = ",";
+    private static final String OPERATION_SUPPLY = "supply";
+    private static final String OPERATION_BUY = "buy";
+    private static final String REPORT_HEADER_SUPPLY = "supply";
+    private static final String REPORT_HEADER_BUY = "buy";
+    private static final String REPORT_HEADER_RESULT = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
```
olekskov, line 17: Please split public method into 3 parts (reading from file, preparing report, writing to file) and extract each part to separate private method.
Then call these 3 methods from public one to make it short and clear for understanding

olekskov (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+        return fileContent;
+    }
+
+    private int processFileContent(List<String> fileContent, String operationType) {
+        int total = 0;
+        for (String line : fileContent) {
+            String[] parts = line.split(DELIMITER);
+            if (parts.length == 2) {
+                String operation = parts[0].trim();
+                int amount = Integer.parseInt(parts[1].trim());
```
Ivan95kos, line 54: use constants instead of numbers

PR #996, opened 2023-06-12
Hw solution

anna-poplavska:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final int OPERATION_TYPE = 0;
+    private static final int AMOUNT = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        WorkWithFile workWithFile = new WorkWithFile();
+        String[] fileInfo = workWithFile.readFile(fromFileName).split(System.lineSeparator());
+        String report = workWithFile.createReport(fileInfo);
+        workWithFile.writeFile(report, toFileName);
```
olekskov, line 20: 1. Why do we create class instance within same class?
2. Please include command .split(System.lineSeparator()) to readFile() so it produces String[] instead of String. OR you may include it to createReport() so it accepts String instead of String[]. Methods of same class should match each other
```suggestion
        String[] fileInfo = readFile(fromFileName);
        String report = createReport(fileInfo);
        writeFile(report, toFileName);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+    }
+
+    private String createReport(String[] fileInfo) {
+        String[] lineInfo;
+        int numberOfSupply = 0;
+        int numberOfBuy = 0;
+        int totalNumber = 0;
+        StringBuilder report = new StringBuilder();
+        for (String info : fileInfo) {
+            lineInfo = info.split(",");
```
olekskov, line 46: Please introduce constant instead of magic value
```suggestion
            lineInfo = info.split(CSV_SEPARATOR);
```

olekskov (changes requested):

aaananas (approved):

PR #997, opened 2023-06-12
homework solution

margosha99:
`readFromFIle()` method try to read data from input csv file by line and throw a RuntimeException if its not possible
`createReport()` method accepts string data, counts sum of operation type and return report in String format
`writeToFile()` method try to write report to file and throw  a RuntimeException if its not possible

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,59 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String CSV_SEPARATOR = ",";
+    private static final String CSV_OPERATION_TYPE = "buy";
+    private static final int CSV_ROW_ELEMENT = 0;
+    private static final int CSV_ROW_DATA_ELEMENT = 1;
```
okuzan, line 13: I'd go for something like AMOUNT_INDEX, OPERATION_INDEX to make it more informative

okuzan (approved):
Wow, you got it on the first try! Keep it up 💪

PR #1000, opened 2023-06-16
Implemented the method by adding three private methods to the class

RhymeWriter:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String PARAMETER_0_SUPPLY = "supply";
+    private static final String PARAMETER_1_BUY = "buy";
+    private static final String PARAMETER_2_RESULT = "result";
```
okuzan, line 13:
```suggestion
    private static final String SUPPLY_OPERATION = "supply";
    private static final String PARAMETER_BUY_OPERATION = "buy";
    private static final String RESULT = "result";
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+        String[] resultData;
+        File incomeData = new File(fromFileName);
+        try (BufferedReader reader = new BufferedReader(new FileReader(incomeData))) {
+            String lineValue = reader.readLine();
+            while (lineValue != null) {
+                stringBuilder.append(lineValue).append(",");
+                System.out.println(lineValue);
+                lineValue = reader.readLine();
+            }
+            resultData = stringBuilder.toString().split(",");
```
okuzan, line 32: strings should be constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Can't open file", e);
```
okuzan, line 34: include filename in the message

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+        String[] parameters = new String[] {PARAMETER_0_SUPPLY, PARAMETER_1_BUY,
+                PARAMETER_2_RESULT};
+        File resultData = new File(toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(resultData))) {
+            for (int i = 0; i < parameters.length; i++) {
+                bufferedWriter.write(parameters[i] + ","
+                        + reportNumbers[i] + System.lineSeparator());
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file", e);
```
okuzan, line 66: same here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
 
+    private void writeReportToFile(String toFileName, int[] reportNumbers) {
+        String[] parameters = new String[] {PARAMETER_0_SUPPLY, PARAMETER_1_BUY,
+                PARAMETER_2_RESULT};
+        File resultData = new File(toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(resultData))) {
+            for (int i = 0; i < parameters.length; i++) {
+                bufferedWriter.write(parameters[i] + ","
+                        + reportNumbers[i] + System.lineSeparator());
+            }
```
okuzan, line 64: In this method, there should not be any logic related to the generation or composition of the report. Only the string that needs to be written to the file should be passed.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+                supply = supply + Integer.parseInt(incomeData [i + 1]);
+                continue;
+            }
+            if (incomeData[i].equals(PARAMETER_1_BUY)) {
+                buy = buy + Integer.parseInt(incomeData[i + 1]);
+            }
+        }
+        result = supply - buy;
+        return new int[] {supply, buy, result};
+    }
```
okuzan, line 54: While returning an array is a valid approach, it could potentially introduce the need for magic numbers in the next method.

Alternatively, a more concise and readable solution would be to generate the report string right within this method, instead of using an array. This would eliminate the need for indexing and improve code readability.

src/main/java/core/basesyntax/WorkWithFile.java
okuzan, line 32: It's up to you, but returning a data as one string would make processing in the next method easier (we would just break by line separator and deal with each line separately)

okuzan (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
RhymeWriter: i do not get it. What you mean? Should i make a StringBuilder a constant field of the class?

PR #1001, opened 2023-06-17
Initial implementation

barbequeue:


src/main/java/core/basesyntax/ReportBuilder.java
```diff
@@ -0,0 +1,46 @@
+package core.basesyntax;
+
+public class ReportBuilder {
```
barbequeue, line 3: I don't like how this class has 2 responsibilities - it parses CSV string and generates another CSV string. Do you think it would be better to split it into two classes - first would just parse CSV string and return parsed data in form of `String[][]`, the second class will generate CSV from this `String[][]` data.

src/main/java/core/basesyntax/ReportBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+    public static final String DEFAULT_CSV_ROW_SEPARATOR = System.lineSeparator();
+    public static final String DEFAULT_CSV_COLUMN_SEPARATOR = ",";
+    public static final String SUPPLY_LABEL = "supply";
+    public static final String BUY_LABEL = "buy";
+    public static final String RESULT_LABEL = "result";
+
+    public String fromCsvString(String str) {
```
Ivan95kos, line 10: do not use such naming(str, stringBuilder), how to guess what it stores?

src/main/java/core/basesyntax/ReportBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+        for (String line : lines) {
+            String[] columns = line.split(DEFAULT_CSV_COLUMN_SEPARATOR);
+
+            if (columns[0].equals(SUPPLY_LABEL)) {
+                supplyTotal += Integer.parseInt(columns[1]);
+            }
+
+            if (columns[0].equals(BUY_LABEL)) {
+                buyTotal += Integer.parseInt(columns[1]);
+            }
```
Ivan95kos, line 29: All magic numbers, strings, symbols in your code should be constants.

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/FileContentReader.java
```diff
@@ -0,0 +1,43 @@
...
+            throw new RuntimeException("Can't read from the file: " + filePath, e);
+        }
+
+        // To remove rightmost DEFAULT_SYSTEM_LINE_SEPARATOR
+        // applied by the last iteration of reading the file content
+        if (fileContentBuilder.length() > 0) {
+            fileContentBuilder.setLength(
+                    fileContentBuilder.length() - DEFAULT_SYSTEM_LINE_SEPARATOR.length()
+            );
+        }
```
okuzan, line 39: instead of this, try to add lineseparator at the beginning

okuzan (approved):

PR #1016, opened 2023-06-26
Implement getStatistics

MurphyEdward:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
+                    buyAmount += amount;
+                }
+            }
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("Cannot find file", e);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        writeResult(supplyAmount, buyAmount, toFileName);
```
olekskov, line 39: 1. Generally it's not a good practice to perform business logic during keeping active connection to file.
Please make reading of full content first then work with content

2. As you will have 3 logical blocks (read file, generate report, write report top file) please make separate private method for each of these blocks and only call them in public method in order to make it short and simple

olekskov (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+    private static final String BUY = "buy";
+    private static final String COMMA = ",";
+    private int supplyAmount;
+    private int buyAmount;
+    private int resultAmount;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String fileContent = readFile(fromFileName);
+        calculateStatistics(fileContent);
+        writeResult(supplyAmount, buyAmount, resultAmount, toFileName);
```
olekskov, line 21: You simultaneously use class variables (that will be available in writeResult() method) and pass them as arguments to method.
Generally, when it's not very complicated, it's better to avoid using class variables as when logic growth you will have lots of them and such situation is harder to maintain.

My recommendation is to make calculateStatistics() method to generate ready to write String and pass to writeResult() method only this String and fileName (and then rename calculateStatistics -> generateReport as it's purpose changes).
But if you like solution with class variables it's also ok, then remove redundant arguments from writeResult method header

olekskov (changes requested):

Ivan95kos (approved):
Great job, with minor comments
For the future, do not write changes in the master branch, create a separate branch for this

PR #1020, opened 2023-06-28
implemented getStatistic method in WorkWithFile

VitaliiLeshchenko:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        String stringReportFromFile = getReportFromFile(fromFileName);
+        try {
+            Files.write(Path.of(toFileName), stringReportFromFile.getBytes());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file: " + toFileName);
+        }
+    }
 
+    private String getReportFromFile(String fromFileName) {
```
liliia-ponomarenko, line 24: This method is overloaded by logic. Better divide logic into three separate methods(read data, generate a report, and write data to file). And after call these methods one by one in getStatistic public method.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
+        StringBuilder stringBuilder = new StringBuilder();
+        Integer supply = 0;
+        Integer buy = 0;
+        try {
+            for (String fileLine : Files.readString(Path.of(fromFileName))
```
liliia-ponomarenko, line 29: better firstly read all data and then process it.

liliia-ponomarenko (changes requested):
Let’s  fix your solution)

src/main/java/core/basesyntax/WorkWithFile.java
VitaliiLeshchenko: done

src/main/java/core/basesyntax/WorkWithFile.java
VitaliiLeshchenko: carry out this part into function readDataFromFile

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+        return stringBuilder.append(SUPPLY + CSV_SEPARATOR)
+                .append(supply)
+                .append(System.lineSeparator())
+                .append(BUY + CSV_SEPARATOR)
+                .append(buy)
+                .append(System.lineSeparator())
+                .append(RESULT + CSV_SEPARATOR)
+                .append(supply - buy)
+                .toString();
+
```
liliia-ponomarenko, line 52:
```suggestion
```

liliia-ponomarenko (approved):

PR #1026, opened 2023-07-03
implemented readTextFromFile method

DmytroV95:


nacenik (changes requested):
Good solution, let's improve it)

Vasul-Pankiv (approved):

PR #1029, opened 2023-07-04
Implement getStatistic(String fromFileName, String toFileName) method 

Dimagaa:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+        int result = supply - buy;
+        return SUPPLY_FIELD + COMMA_SEPARATOR + supply
+                + System.lineSeparator()
+                + BUY_FIELD + COMMA_SEPARATOR + buy
+                + System.lineSeparator()
+                + RESULT_FIELD + COMMA_SEPARATOR + result
+                + System.lineSeparator();
+    }
 
+    private void storeStatistic(String toFileName, String report) {
```
Vasul-Pankiv, line 57: Better to rename this method, name it like writeDataToFile or something like that, cause "store" sound too general

Vasul-Pankiv (approved):
Great solution!

PR #1049, opened 2023-07-07
added read_from_file, create_report, write_to_file methods

Pafmortim:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        String fileContent = readFileContent(fromFileName);
+        String report = createReport(fileContent);
+        writeToFile(toFileName, report);
+    }
+
+    public static String readFileContent(String filePath) {
```
KseniiaMakarova, line 16:
```suggestion
    private String readFileContent(String filePath) {
```

please make helper methods private
and there is no need to make them static

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+        int totalBuy = 0;
 
+        String[] lines = fileContent.split(System.lineSeparator());
+        StringBuilder result = new StringBuilder();
+
+        for (String line : lines) {
+            String[] parts = line.split(",");
+            if (parts.length == 2) {
+                String action = parts[0].trim();
+                int value = Integer.parseInt(parts[1].trim());
```
KseniiaMakarova, line 42:
```suggestion
            String[] parts = line.split(DELIMITER);
            if (parts.length == CORRECT_PARTS_SIZE) {
                String action = parts[ACTION_INDEX].trim();
                int value = Integer.parseInt(parts[VALUE_INDEX].trim());
```

create constants for these values

KseniiaMakarova (changes requested):

andrii-hoienko (approved):

PR #1055, opened 2023-07-07
implemented getStatistic method

RedArtelerist:
- added method for reading info from file
- added method that generates supply/buy report
- added method for writing report to file

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,73 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
 
+public class WorkWithFile {
+    private static final String DELIMITER = ",";
+    private static final int OPERATION_INDEX = 0;
+    private static final int AMOUNT_INDEX = 1;
+    private static final String SUPPLY_TYPE = "supply";
+    private static final String BUY_TYPE = "buy";
+    private static final String RESULT_TYPE = "result";
+
+    public void getStatistic(String fromFileName, String toFileName) {
+        String[] data = readDataFromFile(fromFileName);
+        String report = createReport(data);
+        writeDataToFile(toFileName, report);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+        writeDataToFile(toFileName, report);
+    }
+
+    private String[] readDataFromFile(String filename) {
+        StringBuilder supplyBuyInfo = new StringBuilder();
+
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(filename))) {
+            String line;
+
+            while ((line = bufferedReader.readLine()) != null) {
```
Elena-Bruyako, line 29: You can create the inner variable 
String line = bufferedReader.readLine() 
to avoid creation in the while block and make your code more readable

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+    }
+
+    private String createReport(String[] data) {
+        StringBuilder report = new StringBuilder();
+        int buyAmount = 0;
+        int supplyAmount = 0;
+
+        for (String line : data) {
+            String[] info = line.split(DELIMITER);
+            String operation = info[OPERATION_INDEX];
```
Elena-Bruyako, line 46: Try to rename it to operationType for example

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+                .append(RESULT_TYPE).append(DELIMITER).append(totalAmount);
+
+        return report.toString();
+    }
 
+    private void writeDataToFile(String filename, String report) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(filename))) {
+            bufferedWriter.write(report);
+        } catch (IOException ex) {
+            throw new RuntimeException("Can't write data to " + filename);
```
Elena-Bruyako, line 70: You could also add an exception to get more information in the stack trace 
 throw new RuntimeException("Can't write data to " + filename, ex);
 
 The same for above catch block

Elena-Bruyako (changes requested):

Andriy-dz (approved):
Perfect

PR #1060, opened 2023-07-13
jv-work-with-file-solution

literandltx:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,60 @@
...
+    private static final String BUY = "buy";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String data = getSumUpFromFile(fromFileName);
+
+        try {
+            Files.write(Path.of(toFileName), data.getBytes());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file, ", e);
+        }
```
aaananas, line 22:
```suggestion
    public void getStatistic(String fromFileName, String toFileName) {
        String data = readFromFile(fromFileName);
        String report = createReport(data);
        writeToFile(report, toFileName);
```
you should divide logic into different methods

aaananas (changes requested):

aaananas (approved):

PR #1076, opened 2023-07-30
getStatistic method implemented

katy-mozharovska:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+        File file = new File(fileName);
+
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(file))) {
+            bufferedWriter.write(reportData);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to " + fileName, e);
+        }
+    }
+
+    private String reportDataGenerator(String[] input) {
```
okuzan, line 51: method names shouldn't be noun phrases, but be based on verbs
```suggestion
    private String generateReport(String[] input) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+                line = bufferedReader.readLine();
+            }
+
+            String fileContent = builder.toString();
+
+            if (fileContent.length() != 0) {
+                return fileContent.split(";");
+            } else {
+                return new String[0];
+            }
```
okuzan, line 35: FYI we don't need else when working with returns like that
```suggestion
            if (fileContent.length() == 0) {
                return new String[0];
            }
            return fileContent.split(";");
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read " + fileName, e);
+        }
```
okuzan, line 38:
```suggestion
        List<String> lines = new ArrayList<>();

        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
            String line;
            while ((line = bufferedReader.readLine()) != null) {
                lines.add(line);
            }
        } catch (IOException e) {
            throw new RuntimeException("Can't read " + fileName, e);
        }

        return lines;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+
+        int result = supplyTotal - buyTotal;
+
+        StringBuilder builder = new StringBuilder();
+        builder.append("supply").append(",").append(supplyTotal).append(System.lineSeparator())
+                .append("buy").append(",").append(buyTotal).append(System.lineSeparator())
+                .append("result").append(",").append(result).append(System.lineSeparator());
+        String reportData = builder.toString();
 
+        return reportData;
```
okuzan, line 75:
```suggestion
           return String.format("supply,%d%nbuy,%d%nresult,%d%n", supplyTotal, buyTotal, result);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+
+    private String reportDataGenerator(String[] input) {
+        int supplyTotal = 0;
+        int buyTotal = 0;
+
+        for (int i = 0; i < input.length; i++) {
+            String[] record = input[i].split(",");
+            String key = record[0];
+            int value = Integer.parseInt(record[1]);
+
```
okuzan, line 59: 0, 1 should be class constants

okuzan (changes requested):

okuzan (approved):

PR #1084, opened 2023-08-09
first version, check please

Alex-zzzz:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+    private static String line;
+    private static String tmp = "";
+    private static String[] dataFromFile;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int supplyValue = 0;
+        int buyValue = 0;
+        dataFromFile = readFromFile(fromFileName);
+
+        for (int i = 0; i < dataFromFile.length; i += 2) {
```
overpathz, line 20: What means "2"? Move all magic numbers to constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+            if (dataFromFile[i].equals("supply")) {
+                supplyValue += Integer.valueOf(dataFromFile[i + 1]);
+            } else {
+                buyValue += Integer.valueOf(dataFromFile[i + 1]);
+            }
+        }
+
+        writeToFile(new String[]{"supply," + supplyValue, "buy," + buyValue, "result,"
```
overpathz, line 28: This string values should be constant as well.
Smth like that:
```
private static final String BUY_OPERATION = "buy";
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+import java.io.BufferedWriter;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static String line;
+    private static String tmp = "";
+    private static String[] dataFromFile;
```
overpathz, line 13: There are static variables, that are not related to the object.
It means they are shared and if multiple threads will be executed, they all will work with this data.
So, this data are not isolated.
Each creation of a report should be performed in the method with the corresponding method and local variables.

As static, you should store only constants (operations, indexes of operations in CSV file)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
```
overpathz, line 15:
```
String textFromFile = readFromFile(fromFileName);
String report = createReport(textFromFile);
```
You have to do smth like that.
Each "logic" operation move to separate method.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+            } else {
+                buyValue += Integer.valueOf(dataFromFile[i + 1]);
+            }
+        }
+
+        writeToFile(new String[]{"supply," + supplyValue, "buy," + buyValue, "result,"
+                + (supplyValue - buyValue)}, toFileName);
+    }
+
+    private static String[] readFromFile(String fileName) {
```
overpathz, line 32: This method should not be static

overpathz (changes requested):
Need some fixes to approve.
In general, your flow is okay

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+    private String line;
+    private String tmp = "";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] textFromFile = readFromFile(fromFileName);
+        String[] report = createReport(textFromFile);
+        writeToFile(report,toFileName);
+    }
+
+    public String[] createReport(String[] dataFromFile) {
```
olekskov, line 22:
```suggestion
    private String[] createReport(String[] dataFromFile) {
```

olekskov (approved):
Good job

PR #1085, opened 2023-08-09
Implemented getStatistic method

ystankevych:


liliia-ponomarenko (changes requested):
Interesting solution! Let’s  improve  it ;)

liliia-ponomarenko (approved):
Great 😊

PR #1099, opened 2023-08-15
method implemented

DenisGorda:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,46 @@
...
+                        supplyTotal += amount;
+                    } else if (operation.equals("buy")) {
+                        buyTotal += amount;
+                    }
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Error occured while reading data from: " + fromFileName, e);
+        } finally {
+            writeReport(toFileName, supplyTotal, buyTotal);
```
Andriy-dz, line 33: this is not very correct behavior, because if we encounter an error when reading from a file, then we will send invalid data to this method and they will be written to the file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,46 @@
...
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] parts = line.split(",");
+                if (parts.length == 2) {
+                    String operation = parts[0].trim();
+                    int amount = Integer.parseInt(parts[1].trim());
+
+                    if (operation.equals("supply")) {
```
Andriy-dz, line 23: let's make 'supply' a constant. Likewise in similar cases

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,46 @@
...
+                        buyTotal += amount;
+                    }
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Error occured while reading data from: " + fromFileName, e);
+        } finally {
+            writeReport(toFileName, supplyTotal, buyTotal);
+        }
+    }
```
Andriy-dz, line 35: let's add two more private methods readFromFile and calculateRepost - which will count and compile all the information into a String. And in the getStatistic method, you just need to call the corresponding methods one by one

Andriy-dz (changes requested):
Good job! Let's improve your solution)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+        String report = calculateReport(readFromFile(fromFileName));
+        writeReport(toFileName, report);
+    }
+
+    private String readFromFile(String fileName) {
+        StringBuilder data = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                data.append(line).append("\n");
```
Andriy-dz, line 23: better to use System.lineSeparator()

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+        return data.toString();
+    }
+
+    private String calculateReport(String data) {
+        int supplyTotal = 0;
+        int buyTotal = 0;
+
+        String[] lines = data.split("\n");
+        for (String line : lines) {
+            String[] parts = line.split(",");
```
Andriy-dz, line 37: create constant for comma

Andriy-dz (approved):
Good Job!

PR #1105, opened 2023-08-16
Jv-work-with-file task solution

mkazarian-1:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,43 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final int SPLIT_LINE_STATUS = 0;
+    private static final int SPLIT_LINE_VALUE = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
```
Rommelua, line 13: image

Rommelua (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
 public class WorkWithFile {
+    private static final int SPLIT_LINE_STATUS = 0;
+    private static final int SPLIT_LINE_VALUE = 1;
+    private static final int SUPPLY = 0;
+    private static final int BUY = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int[] supplyAndBuy = {0, 0};
+        readFromFile(fromFileName, supplyAndBuy);
+        writerToFile(toFileName, supplyAndBuy);
```
Rommelua, line 18:
```suggestion
        String dataFromFile = readFile(fromFileName);
        String report = createReport(dataFromFile);
        writeToFile(report, toFileName);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
+    }
 
+    private void writerToFile(String file, int[] supplyAndBuy) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(file))) {
+            writer.write("supply," + supplyAndBuy[SUPPLY] + System.lineSeparator()
+                    + "buy," + supplyAndBuy[BUY] + System.lineSeparator()
+                    + "result," + (supplyAndBuy[SUPPLY] - supplyAndBuy[BUY])
+                    + System.lineSeparator());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file ", e);
```
Rommelua, line 48:
```suggestion
            throw new RuntimeException("Can't write data to file " + file, e);
```

Rommelua (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
mkazarian-1: If I have understood and implemented everything correctly, your implementation is better in terms of clean code, but it seems to me that this implementation will have memory problems with large amounts of data, since instead of reading information line by line, we read and store the entire block of information at once.

Rommelua (approved):

mkazarian-1:
So how is right or wrong? "If I have understood and implemented everything correctly, your implementation is better in terms of clean code, but it seems to me that this implementation will have memory problems with large amounts of data, since instead of reading information line by line, we read and store the entire block of information at once."

PR #1109, opened 2023-08-16
work with file

KDanil2000:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+    private static final String SUPPLY_STRING = "supply";
+    private static final String BUY_STRING = "buy";
+    private static final String RESULT_STRING = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        File fromFile = new File(fromFileName);
+        File toFile = new File(toFileName);
+        String text = readFile(fromFile);
+        String statistic = getStatisticFromText(text);
+        writeStatistic(toFile, statistic);
```
NazarSmith228, line 22: you have weird method ordering - since you're invoking `readFile` first, then it should be the first method after the current one

and so on

NazarSmith228 (changes requested):
Good job! Now it's time to improve your solution

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+        for (String line : lines) {
+            String[] operationInfo = line.split(SEPARATOR_STRING);
+            int quantity = Integer.parseInt(operationInfo[1]);
+            if (operationInfo[0].equals(SUPPLY_STRING)) {
+                supplySum += quantity;
+            } else {
+                buySum += quantity;
+            }
+        }
+        int difference = supplySum - buySum;
```
Elena-Bruyako, line 44: What does it mean difference? difference from what? 
bad naming - result will be better, and StringBuilder could be - report

Elena-Bruyako (approved):

PR #1110, opened 2023-08-16
Added solution

AlexRosenkreuz:
Implemented getStatistic() method which calls readFromFile(String fileName), calculateReport() and writeToFile() methods to create .csv report.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,70 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
 
+public class WorkWithFile {
+    private static final String COMMA = ",";
+    private static final String BUY = "buy";
+    private static final String SUPPLY = "supply";
+    private static final String RESULT = "result";
+
+    public void getStatistic(String fromFileName, String toFileName) {
+        try {
+            String report = calculateReport(readFromFile(fromFileName));
+            writeToFile(toFileName, report);
+            System.out.println("Report generated successfully.");
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    private String readFromFile(String fileName) throws IOException {
+        StringBuilder data = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                data.append(line).append(System.lineSeparator());
+            }
+        }
+        return data.toString();
+    }
+
+    private String calculateReport(String data) {
+        int supplyTotal = 0;
+        int buyTotal = 0;
+
+        String[] lines = data.split(System.lineSeparator());
+        for (String line : lines) {
+            String[] parts = line.split(COMMA);
+            if (parts.length == 2) {
+                String operationType = parts[0].trim();
+                int amount = Integer.parseInt(parts[1].trim());
+                if (operationType.equalsIgnoreCase(SUPPLY)) {
+                    supplyTotal += amount;
+                } else if (operationType.equalsIgnoreCase(BUY)) {
+                    buyTotal += amount;
+                }
+            }
+        }
+
+        int result = supplyTotal - buyTotal;
+
+        StringBuilder report = new StringBuilder();
+        report.append(SUPPLY + COMMA).append(supplyTotal).append(System.lineSeparator());
+        report.append(BUY + COMMA).append(buyTotal).append(System.lineSeparator());
+        report.append(RESULT + COMMA).append(result);
+
+        return report.toString();
+    }
+
+    private void writeToFile(String fileName, String content) throws IOException {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(content);
+        }
     }
 }
+


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+    private static final String RESULT = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        try {
+            String report = calculateReport(readFromFile(fromFileName));
+            writeToFile(toFileName, report);
+            System.out.println("Report generated successfully.");
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
```
NazarSmith228, line 22: you should better handle corresponding exceptions in `readFromFile` and `writeToFile` methods

this ways you'll be able to throw dedicated exceptions from there - you should not just print stack trace

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+        return data.toString();
+    }
+
+    private String calculateReport(String data) {
+        int supplyTotal = 0;
+        int buyTotal = 0;
 
+        String[] lines = data.split(System.lineSeparator());
+        for (String line : lines) {
+            String[] parts = line.split(COMMA);
```
NazarSmith228, line 42: I would rename this to `operationInfo` - gives more context in contrast to `parts`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+            if (parts.length == 2) {
+                String operationType = parts[0].trim();
+                int amount = Integer.parseInt(parts[1].trim());
```
NazarSmith228, line 45: you don't need `trim()` - assume that the input file is valid

also I would rename `amount` to `operationAmount` - gives more context

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+                if (operationType.equalsIgnoreCase(SUPPLY)) {
+                    supplyTotal += amount;
+                } else if (operationType.equalsIgnoreCase(BUY)) {
```
NazarSmith228, line 48: you can use simple `else` here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+                }
+            }
+        }
+
+        int result = supplyTotal - buyTotal;
+
+        StringBuilder report = new StringBuilder();
+        report.append(SUPPLY + COMMA).append(supplyTotal).append(System.lineSeparator());
+        report.append(BUY + COMMA).append(buyTotal).append(System.lineSeparator());
+        report.append(RESULT + COMMA).append(result);
```
NazarSmith228, line 59: there is no need to use String concatenation - it's worse in terms of memory

you can use additional `append()`

NazarSmith228 (changes requested):
Great job! Now it's time to improve your solution

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,70 @@
...
+    private static final String SUPPLY = "supply";
+    private static final String RESULT = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        try {
+            String report = calculateReport(readFromFile(fromFileName));
+            writeToFile(toFileName, report);
+            System.out.println("Report generated successfully.");
+        } catch (IOException e) {
+            e.printStackTrace();
```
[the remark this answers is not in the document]
AlexRosenkreuz: Moved catch blocks to those methods and added error messages.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+
+    private String readFromFile(String fileName) {
+        StringBuilder data = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                data.append(line).append(System.lineSeparator());
+            }
+        } catch (IOException e) {
+            System.out.println("Reading file failed.");
```
NazarSmith228, line 29: you should better re-throw a RuntimeException with a message like `Failed to read from file:` + filename and pass the caught exception as well

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+        }
+        return data.toString();
+    }
 
+    private String calculateReport(String data) {
+        int supplyTotal = 0;
+        int buyTotal = 0;
+
```
NazarSmith228, line 37: redundant blank line

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+        report.append(RESULT).append(COMMA).append(result);
+
+        return report.toString();
+    }
+
+    private void writeToFile(String fileName, String content) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(content);
+        } catch (IOException e) {
+            System.out.println("Writing file failed.");
```
NazarSmith228, line 62: you should better re-throw a RuntimeException with a message like `Failed to write to file:` + filename and pass the caught exception as well

NazarSmith228 (changes requested):
Great job! Just some minor remarks left to fix

Elena-Bruyako (approved):

PR #1112, opened 2023-08-16
impelement methodes: getFileStat, getWrittenFile, getStatistic

anigilator479:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+    private static final String SPECIAL_CHAR = ",";
```
NazarSmith228, line 11: this is not a special char, it's just a `comma` or `separator`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+
+    private StringBuilder getFileStat(String fromFileName) {
+        File fileToRead = new File(fromFileName);
```
NazarSmith228, line 16: you can use filename directly in `FileReader`  and `FileWriter` - no need to create `File` instances

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+    private static final String SPECIAL_CHAR = ",";
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+
+    private StringBuilder getFileStat(String fromFileName) {
```
NazarSmith228, line 15: this method is too complex and large

you should have 3 separate methods:

1. Read the file data and return it (`String` or `String[]` should be fine)
2. Calculate the report based on the data you read in the first method
3. Write the report to a file

Currently this method reads the data and calculate the report at the same time

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+            stringBuilder.append("supply," + supply + System.lineSeparator())
+                    .append("buy," + buy + System.lineSeparator())
+                    .append("result," + result + System.lineSeparator());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read this file", e);
+        }
+        return stringBuilder;
+    }
 
+    private void getWrittenFile(StringBuilder stringBuilder, String toFileName) {
```
NazarSmith228, line 44: pretty much the same with this method - it's too complex and large, try to simplify it, remove everything unnecessary

also the method name is confusing - `void` return type and `get` in method name :)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+                bufferedWriter.write(stat + System.lineSeparator());
+            } catch (IOException e) {
+                throw new RuntimeException("Can't write this file",e);
+            }
+        }
+    }
+
+    public void getStatistic(String fromFileName, String toFileName) {
+        StringBuilder stringBuilder = getFileStat(fromFileName);
+        getWrittenFile(stringBuilder,toFileName);
```
NazarSmith228, line 67: you should't declare a new method - use existing one

NazarSmith228 (changes requested):
Good job, now let's improve your solution

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,91 @@
...
+        return true;
+    }
+
+    private void fileExistCheck(File toFileName) {
+        if (toFileName.exists()) {
+            toFileName.delete();
+        }
+    }
 
+    private void makeWrittenFile(StringBuilder stringBuilder, File toFileName) {
```
Elena-Bruyako, line 78: The name of method doesn't describe what it actually do. Also StringBuilder stringBuilder parameter is bad naming, it doesn't give an information what exactly you provide in this method.

```suggestion
   private void writeToFile(String data, String toFileName) {
        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
            bufferedWriter.write(data);
        } catch (IOException e) {
            throw new RuntimeException("Can`t write data to the file " + toFileName, e);
        }
```

Elena-Bruyako (changes requested):
Make your code more readable, pay attention to [common mistakes ](https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file#common-mistakes), think about the naming of methods and variables

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,82 @@
...
+        }
+        return stringBuilder.toString().split(System.lineSeparator());
+    }
+
+    private StringBuilder createReport(String[] text) {
+        int supply = 0;
+        int buy = 0;
+        int result;
+        String[] strings;
+        for (String s : text) {
```
Elena-Bruyako, line 38: s - it's bad naming. You should use understandable names

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,82 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file" + fromFileName, e);
+        }
+        return stringBuilder.toString().split(System.lineSeparator());
+    }
+
+    private StringBuilder createReport(String[] text) {
+        int supply = 0;
+        int buy = 0;
+        int result;
```
Elena-Bruyako, line 36: It's a bad practice to declare the variable not in place where you are going to use it

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,82 @@
...
+        }
+        result = supply - buy;
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("supply,").append(supply).append(System.lineSeparator())
+                .append("buy,").append(buy).append(System.lineSeparator())
+                .append("result,").append(result).append(System.lineSeparator());
+        return stringBuilder;
+    }
+
+    private boolean createFile(File toFileName) {
```
Elena-Bruyako, line 54: you don't need to create file and checking if file exists as you are using BufferedWriter

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,82 @@
...
+            strings = s.split(SEPARATOR);
+            if (strings[0].equals(SUPPLY)) {
+                supply += Integer.valueOf(strings[1]);
+            } else if (strings[0].equals(BUY)) {
+                buy += Integer.valueOf(strings[1]);
+            }
+        }
+        result = supply - buy;
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("supply,").append(supply).append(System.lineSeparator())
```
Elena-Bruyako, line 48: use constants SUPPLY and create COMMA constant 
the same for buy and result

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,82 @@
...
+                .append("buy,").append(buy).append(System.lineSeparator())
+                .append("result,").append(result).append(System.lineSeparator());
+        return stringBuilder;
```
Elena-Bruyako, line 51: Return StringBuilder isn't good idea. You method createReport should do all logic that related to create some report, so the method should return completed result
return stringBuilder.toString();

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,82 @@
...
+    }
+
+    private void fileExistCheck(File toFileName) {
+        if (toFileName.exists()) {
+            toFileName.delete();
+        }
+    }
 
+    private void writeToFile(StringBuilder statistic, File toFileName) {
+        String[] stats = statistic.toString().split(System.lineSeparator());
```
Elena-Bruyako, line 70: why did you do this?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,82 @@
...
+        fileExistCheck(toFileName);
+        createFile(toFileName);
+        for (String stat : stats) {
+            try (BufferedWriter bufferedWriter =
+                         new BufferedWriter(new FileWriter(toFileName, true))) {
+                bufferedWriter.write(stat + System.lineSeparator());
```
Elena-Bruyako, line 76: in this method you need to get completed report and just write it to the file, don't need to change or add something

Elena-Bruyako (changes requested):

liliia-ponomarenko (approved):
Great job 👏

PR #1115, opened 2023-08-16
implement file formating, and creation methods

Parashchuk:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+            StringBuilder parsedCsv = new StringBuilder();
+            BufferedReader reader = new BufferedReader(fileReader);
+
+            String value;
+            if ((value = reader.readLine()) == null) {
+                return null;
+            }
+            do {
+                parsedCsv.append(value).append(PARSED_STRING_SEPARATOR);
+            } while ((value = reader.readLine()) != null);
```
liliia-ponomarenko, line 32: If you use while you can avoid checking

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] data = readDataFromFile(fromFileName);
+        String report = formReport(data);
+        createNewDataFile(toFileName, report);
+    }
+
+    public String[] readDataFromFile(String fromFileName) {
+        try (FileReader fileReader = new FileReader(fromFileName)) {
+            StringBuilder parsedCsv = new StringBuilder();
+            BufferedReader reader = new BufferedReader(fileReader);
```
liliia-ponomarenko, line 24:
```suggestion
         StringBuilder parsedCsv = new StringBuilder();
        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
```
1) you should keep the least amount of logic in the try block that can cause an exception 
2) BufferedReader extends from Reader that implements Closeable. It’s mean that you should  use it in a try-with-resources block

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+            if ((value = reader.readLine()) == null) {
+                return null;
+            }
+            do {
+                parsedCsv.append(value).append(PARSED_STRING_SEPARATOR);
+            } while ((value = reader.readLine()) != null);
+
+            return parsedCsv.toString().split(PARSED_STRING_SEPARATOR);
+        } catch (IOException e) {
+            throw new RuntimeException("Try to read file was unsuccessful", e);
```
liliia-ponomarenko, line 36: make this message more informative. It will be great to add an input parameter (fromFileName) to the message.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+        }
+    }
+
+    private String formReport(String[] parsedString) {
+        int buyAmount = 0;
+        int supplyAmount = 0;
+        StringBuilder result = new StringBuilder();
+
+        for (String data: parsedString) {
+            int amountNumber = Integer.parseInt(data.substring(data.indexOf(',') + 1));
```
liliia-ponomarenko, line 46: better use `split(",")`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
+                .append(FORMED_STRING_SEPARATOR)
+                .append(supplyAmount - buyAmount)
+                .toString();
+    }
 
+    private void createNewDataFile(String fileName, String data) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(fileName))) {
+            bufferedWriter.write(data);
+        } catch (IOException e) {
+            throw new RuntimeException("Try to write file was unsuccessful", e);
```
liliia-ponomarenko, line 73: the same about the message

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,76 @@
...
 public class WorkWithFile {
+    private static final String PARSED_STRING_SEPARATOR = ";";
+    private static final String FORMED_STRING_SEPARATOR = ",";
+    private static final String BUY_CATEGORY_NAME = "buy";
+    private static final String SUPPLY_CATEGORY_NAME = "supply";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] data = readDataFromFile(fromFileName);
+        String report = formReport(data);
+        createNewDataFile(toFileName, report);
```
liliia-ponomarenko, line 18:
```suggestion
        writeFile(toFileName, report);
```

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+        int supplyAmount = 0;
+        StringBuilder result = new StringBuilder();
+
+        for (String data: parsedString) {
+            int amountNumber = Integer.parseInt(data.split(",")[1]);
+            if (data.contains(BUY_CATEGORY_NAME)) {
+                buyAmount += amountNumber;
+            } else if (data.contains(SUPPLY_CATEGORY_NAME)) {
+                supplyAmount += amountNumber;
+            }
```
liliia-ponomarenko, line 46:
```suggestion
            String records = data.split(FORMED_STRING_SEPARATOR);
            int amountNumber = Integer.parseInt(records[OPERATION_AMOUNT_INDEX]);
            if (records[OPERATION_INDEX].equels(BUY_CATEGORY_NAME)) {
                buyAmount += amountNumber;
            } else  {
                supplyAmount += amountNumber;
            }
```
I meant to do something like this. It increases productivity in your solution (why?)

liliia-ponomarenko (changes requested):
Great but let’s fix one thing ;)

src/main/java/core/basesyntax/WorkWithFile.java
Parashchuk: Increases readability - no magic numbers

liliia-ponomarenko (approved):
Great 😊

PR #1117, opened 2023-08-17
Implemented method getStatistic(), which return data from market.

pastcwdt:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+                buy += amount;
+            } else {
+                supply += amount;
+            }
+        }
+        int result = supply - buy;
+        report.append(SUPPLY_OPERATION).append(COMMA).append(supply).append(System.lineSeparator())
+                .append(BUY_OPERATION).append(COMMA).append(buy).append(System.lineSeparator())
+                .append(RESULT).append(COMMA).append(result);
+        return report.toString();
```
liliia-ponomarenko, line 56:
```suggestion
        return report.append(SUPPLY_OPERATION).append(COMMA).append(supply).append(System.lineSeparator())
                .append(BUY_OPERATION).append(COMMA).append(buy).append(System.lineSeparator())
                .append(RESULT).append(COMMA).append(result).toString();
```
Can be simplified

liliia-ponomarenko (approved):
Great 😊

PR #1119, opened 2023-08-17
report

DenisBogusevich:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private final char separator = ',';
+    private String supply = "supply";
+    private String buy = "buy";
+    private String result = "result";
```
liliia-ponomarenko, line 13: Let’s  make these variables constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private final char separator = ',';
```
liliia-ponomarenko, line 10:
```suggestion
    private static final String COMA_SEPARATOR = ",";
```
avoid using `String.valueOf()`

liliia-ponomarenko (changes requested):
Good job! Let’s fix a few mistakes  ;)

liliia-ponomarenko (approved):
LGTM 🔥

PR #1125, opened 2023-08-21
hw-jv-work-with-file

IhorTrokhymchuk:


okuzan (changes requested):
image
Please divide the logic into three methods - reading data, generating a report from said data, writing a report to a file
All literals (at least indexes) must be constants. When throwing an exception, include the filename in the message

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
kozhukhovsky, line 10: This method should ideally only consist of calls to other methods. For example:

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+
+            String[] stringInput = readString.split("\\W+");
+            for (int i = 0; i < stringInput.length / 2; i++) {
+                if (stringInput[i * 2].equals("supply")) {
+                    supplySum += Integer.parseInt(stringInput[i * 2 + 1]);
+                } else if (stringInput[i * 2].equals("buy")) {
+                    buySum += Integer.parseInt(stringInput[i * 2 + 1]);
+                }
+            }
+            String statistic = generateStatistic(supplySum, buySum);
```
```suggestion
        if (!readString.isEmpty()) {
            int supplySum = 0;
            int buySum = 0;

            String[] stringInput = readString.split("\\W+");
            for (int i = 0; i < stringInput.length / 2; i++) {
                if (stringInput[i * 2].equals("supply")) {
                    supplySum += Integer.parseInt(stringInput[i * 2 + 1]);
                } else if (stringInput[i * 2].equals("buy")) {
                    buySum += Integer.parseInt(stringInput[i * 2 + 1]);
                }
            }
            String statistic = generateStatistic(supplySum, buySum);
```

move this to the createReport method

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        String readString = readFile(fromFileName);
+        int[] sums = createReport(readString);
+        String statistic = generateStatistic(sums[0], sums[1]);
```
kozhukhovsky, line 13: generateStatistic should be inside createReport

kozhukhovsky, line 13: all indices must be converted to constants, since I don’t understand what exactly is under this index

kozhukhovsky (changes requested):
There's still a little bit left :)

kozhukhovsky (approved):

PR #1131, opened 2023-08-31
implemented the getStatistic method to create a report

Petro-Smoliar:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+    private int[] calculateSupplyAndBuy(String fromFileName) {
+        int supply = 0;
+        int buy = 0;
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
+            String line = reader.readLine();
+            while (line != null) {
+                if (line.split(",")[INDEX_OPERATION].equals("supply")) {
+                    supply += Integer.parseInt(line.split(",")[INDEX_QUANTITY]);
+                } else {
+                    buy += Integer.parseInt(line.split(",")[INDEX_QUANTITY]);
```
okuzan, line 36: extract common part from if-else

okuzan (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+                lines.add(line);
+                line = reader.readLine();
+            }
+            return lines.toArray(new String[0]);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private int[] calculateSupplyAndBuy(String[] lines) {
```
kozhukhovsky, line 38: All OK. Just need to change this method a bit.
Your report is generated in the writeToFile method, but this is not entirely correct. The task of that method is simply to take the data and write it to a file. The data must already be generated.
Therefore, processing
"supply," + supply + System.lineSeparator()
                     + "buy," + buy + System.lineSeparator()
                     + "result," + result)
must be added to the createReport() method

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+    private static final int INDEX_QUANTITY = 1;
+    private static final String DELIMITER = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] lines = readingFile(fromFileName);
+        int[] calculate = calculateSupplyAndBuy(lines);
+        writingFile(toFileName, calculate[INDEX_SUPPLY], calculate[INDEX_BUY]);
+    }
+
+    private String[] readingFile(String fromFileName) {
```
```suggestion
    private String[] readFile(String fromFileName) {
```

src/main/java/core/basesyntax/WorkWithFile.java
kozhukhovsky, line 38:
```suggestion
    private String createReport(String[] lines) {
```

Even better, this method should already return a String.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+            if (operation.equals("supply")) {
+                supply += quantity;
+            } else {
+                buy += quantity;
+            }
+        }
+        return new int[]{supply, buy};
+    }
 
+    private void writingFile(String toFileName, int supply, int buy) {
```
kozhukhovsky, line 54:
```suggestion
    private void writeFile(String toFileName, String data) {
```

kozhukhovsky (changes requested):

kozhukhovsky (approved):

PR #1134, opened 2023-09-03
implemented a solution

breakingtumble:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,51 @@
...
+
 public class WorkWithFile {
+    private static final int FIRST_COLUMN_INDEX = 0;
+    private static final int SECOND_COLUMN_INDEX = 1;
+    private static final String SEPARATOR = ",";
+    private static final String BUY_WORD_FOR_COMPARING = "buy";
+    private static final String SUPPLY_WORD_FOR_COMPARING = "supply";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
```
sokimaaa, line 17: Divide program logic : Large methods is very difficult to read and understand. So methods should be short, between 5-15 lines. And it’s important for you to remember that one method should be responsible for one task.

Examples of dividing: 
1) Method responsible for reading from file 
2) Method responsible for making a report
3) Method responsible for writing to file

sokimaaa (changes requested):

PR #1136, opened 2023-09-05
created

KhrystynaDidykh:


kozhukhovsky (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+        }
 
+        StringBuilder builder = new StringBuilder();
+        builder.append("supply," + supplyCount + System.lineSeparator()
+                + "buy," + buyCount + System.lineSeparator()
+                + "result," + result + System.lineSeparator());
+        return builder.toString();
     }
+
+    public void writeReportToFile(String fileData, String toFileName) {
```
kozhukhovsky, line 48:
```suggestion
    private void writeReportToFile(String fileData, String toFileName) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+            while ((text = reader.readLine()) != null) {
+                builder.append(text).append(System.lineSeparator());
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return builder.toString();
+    }
+
+    public String createReport(String data) {
```
kozhukhovsky, line 26:
```suggestion
    private String createReport(String data) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+    static final int ZERO = 0;
+    static final int ONE = 1;
+
+    public String readFile(String fromFileName) {
```
kozhukhovsky, line 13:
```suggestion
    private String readFile(String fromFileName) {
```

kozhukhovsky (changes requested):
Service methods are best kept private.
It is necessary to place public methods first, then all private ones.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+                + "buy," + buyCount + System.lineSeparator()
+                + "result," + result + System.lineSeparator());
+        return builder.toString();
+    }
 
+    private void writeReportToFile(String fileData, String toFileName) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName))) {
+            writer.write(fileData);
+            writer.flush();
+            writer.close();
```
okuzan, line 58: we use try with resources so no need to care about it
```suggestion
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
okuzan, line 60: image

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+        return builder.toString();
+    }
+
+    private String createReport(String data) {
+        int supplyCount = 0;
+        int buyCount = 0;
+        int result = 0;
+        String[] lines = data.split(System.lineSeparator());
+        for (String line : lines) {
+            String[] part = line.split(",");
```
okuzan, line 38: make this SEPARATOR/DELIMITER constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    static final int ZERO = 0;
+    static final int ONE = 1;
```
okuzan, line 11: don't name constants after their values, they may change - name them after the role they serve in the app. It's the opportunity to increase readability - tell to those who will read your code, what's going on
OPERATION_INDEX, AMOUNT_INDEX

okuzan (changes requested):

olekskov (approved):
Good job

PR #1143, opened 2023-09-13
Roma Menzheres's realisation of "Work with csv file" task

RomanMenzheres:
Added realisation for getStatistic method


olekskov (approved):
Good job

PR #1150, opened 2023-09-17
Implement method getStatistic(String fromFileName, String toFileName)

andriimikhaylovskiy:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        File inputFile = new File(fromFileName);
+        String[] words = readFromFile(inputFile).split("\\W+");
```
okuzan, line 17: extract regex to a constant

okuzan (changes requested):
Before we proceed with the further review, please make sure getstatistic method contains only three method calls, and no complex login in itself - delegate reading data from file, generating a report, and writing results to a file to 3 separate private utility methods

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+        }
+
+        String report = generateReport(supplySum, buySum);
+        File outputFile = new File(toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(outputFile))) {
+            bufferedWriter.write(report);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to create a file: " + outputFile.getName(), e);
+        }
+    }
```
olekskov, line 38: Logic of method is still overwhelmed.
Please extract part to new method like writeToFile() and other parts move to existing generateReport() and readFromFile() so public method look like below:
```suggestion
        String fileContent = readFromFile(fromFileName);
        String report = generateReport(fileContent);
        writeToFile(report, toFileName);
    }
```

olekskov (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final String WORD_SPLIT_REGEX = "\\W+";
+
     public void getStatistic(String fromFileName, String toFileName) {
```
Ivan95kos, line 16: this method should contain three calls of other methods: readFile() - returns content, generateReport() -(you pass contents here), writeToFile()-(you pass report and filename here ), this will be a good code decomposition. example:
```
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+        }
+
+        String report = generateReport(supplySum, buySum);
+        writeToFile(toFileName, report);
+    }
+
+    private String generateReport(int supplySum, int buySum) {
+        int resultSum = supplySum - buySum;
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append(SUPPLY).append(",").append(supplySum).append(System.lineSeparator())
```
Ivan95kos, line 38: All magic numbers, strings, and symbols in your code should be constants.
```suggestion
        stringBuilder.append(SUPPLY).append(SEPARATOR).append(supplySum).append(System.lineSeparator())
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
 
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
```
Ivan95kos, line 11: The name of the constant should not be the same as the value being stored

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,72 @@
...
+        int supplySum = 0;
+        int buySum = 0;
+
+        for (int i = 0; i < words.length; i += 2) {
+            if (words[i].equals(SUPPLY_CONSTANT)) {
+                supplySum += Integer.parseInt(words[i + 1]);
+            } else if (words[i].equals(BUY_CONSTANT)) {
+                buySum += Integer.parseInt(words[i + 1]);
+            }
+        }
```
okuzan, line 34: this is a fragile approach, a more robust one would be to have constants for indexes in the array by which you can get values, OPERATION_INDEX, AMOUNT_INDEX, but overall it's ok as is

okuzan (approved):
Good job

PR #1152, opened 2023-09-19
Added soulution

slaybrute:


okuzan (changes requested):
The getStatistic method should delegate all the logic to private utility methods - one for reading info into a variable, the second for generating a report, and the third for writing said report to a file. Then getStatistic becomes a high-level method coordinating the process, while each utility method encapsulates specific logic, thus adhering to the Single Responsibility Principle. This is called decoupling

slaybrute:
Hence, I ought to change access modifiers of public ArrayList<String[]> getDataFromFile and public String getStringWithStatistic to private for incapsulating behavior from users. We only should have access to derived result in getStatistic method.

Ivan95kos (changes requested):
Good job overall!

Ivan95kos (approved):
Nice

PR #1156, opened 2023-09-27
realized WorkWithFile class

fedorovychh:
realized getStatistic method
implemetned readFile method
implemented calculateReport method

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -7,7 +7,9 @@ import java.io.FileWriter;
 import java.io.IOException;
 
 public class WorkWithFile {
-    private static final int DEFAULT_OPERATIONS_NUMBER = 2;
+    private static final int ASCII_NEGATIVE_INDEX = -1;
+    private static final int OPERATION_COLUMN = 0;
+    private static final int MONEY_COLUMN = 1;
     private static final String FIRST_OPERATION_NAME = "supply";
     private static final String SECOND_OPERATION_NAME = "buy";
     private static final String RESULT_OPERATION_NAME = "result";
@@ -26,7 +28,7 @@ public class WorkWithFile {
         StringBuilder readFileBuilder = new StringBuilder();
         try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
             int charTextValue = bufferedReader.read();
-            while (charTextValue != -1) {
+            while (charTextValue != ASCII_NEGATIVE_INDEX) {
                 readFileBuilder.append((char) charTextValue);
                 charTextValue = bufferedReader.read();
             }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
 public class WorkWithFile {
+    private static final int DEFAULT_OPERATIONS_NUMBER = 2;
+    private static final String FIRST_OPERATION_NAME = "supply";
+    private static final String SECOND_OPERATION_NAME = "buy";
+    private static final String RESULT_OPERATION_NAME = "result";
+    private static final String SYMBOL_COMMA = ",";
+    private static final String SYMBOL_LINE_SEPARATOR = "\n";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        File myFile = new File(toFileName);
```
mnyshenko, line 19: You're creating a new file too early, what if there is an exception when reading from a file? And the whole time you're reading from the file you're holding an object that is stored in myFile variable, which is insufficient from a memory usage perspective

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(myFile))) {
+            bufferedWriter.write(calculateReport(readFile(fromFileName)));
```
mnyshenko, line 21: calculateReport() and readFile() don't seem to be able to produce IOException so there is no point in putting them inside the try-catch

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+    private static final int DEFAULT_OPERATIONS_NUMBER = 2;
+    private static final String FIRST_OPERATION_NAME = "supply";
+    private static final String SECOND_OPERATION_NAME = "buy";
+    private static final String RESULT_OPERATION_NAME = "result";
+    private static final String SYMBOL_COMMA = ",";
+    private static final String SYMBOL_LINE_SEPARATOR = "\n";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        File myFile = new File(toFileName);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(myFile))) {
```
mnyshenko, line 20: I have an assumption that there is no need to create a File object, you can just a file name here in FileWriter as you did with FileReader in readFile() method.

mnyshenko (changes requested):
Just a few points to consider and refactor

src/main/java/core/basesyntax/WorkWithFile.java
proxychang: It is necessary to add try/catch because both new FileWriter and write method can potencially produce an exception

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+            throw new RuntimeException("Cannot read data from the file " + fromFileName, e);
+        }
+    }
 
+    private String calculateReport(String readFile) {
+        int supply = 0;
+        int buy = 0;
+        int parsedValue;
+        String[] separatedText = readFile.split(SYMBOL_LINE_SEPARATOR);
+        String[][] statisticTable = new String[separatedText.length][DEFAULT_OPERATIONS_NUMBER];
```
d1sam, line 44: Do you really need this `String[][] statisticTable`. I think that you can write logic of money counting without this array.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+        for (int i = 0; i < separatedText.length; i++) {
+            for (int j = 0; j < DEFAULT_OPERATIONS_NUMBER; j++) {
+                statisticTable[i][j] = separatedText[i].split(SYMBOL_COMMA)[j];
+            }
+        }
```
d1sam, line 49: When you remove String[][] statisticTable you will be not needed in this `for`-loop

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Cannot write data to the file " + toFileName, e);
+        }
+    }
+
+    private String readFile(String fromFileName) {
+        StringBuilder readFileBuilder = new StringBuilder();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            int charTextValue = bufferedReader.read();
+            while (charTextValue != -1) {
```
d1sam, line 29: Try to avoid magic-numbers.

d1sam (changes requested):
Good job, but there are still some mistakes need to be resolved. If you have any questions please write in a slack :)

proxychang:
> Good job, but there are still some mistakes need to be resolved. If you have any questions please write in a slack :)

it is done!

d1sam (approved):
Well done.

PR #1157, opened 2023-09-28
Added logic to getStatistic method

HovorukhaBohdan:


src/main/java/core/basesyntax/WorkWithFile.java
d1sam, line 1: Divide program logic

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,38 @@
...
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
+                BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName))) {
+            String values = reader.readLine();
+            int totalSupply = 0;
+            int totalBuy = 0;
+
+            while (values != null) {
+                String[] value = values.split(",");
```
d1sam, line 18: Try to avoid magic-numbers and `String` literals, make them as a constants.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,38 @@
...
+                }
+
+                values = reader.readLine();
+            }
+
+            int result = totalSupply - totalBuy;
 
+            writer.write("supply," + totalSupply + System.lineSeparator()
+                    + "buy," + totalBuy + System.lineSeparator()
+                    + "result," + result);
```
d1sam, line 33: Try to avoid magic-numbers and `String` literals, make them as a constants.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,38 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
d1sam, line 35: Better re-throw RuntimeException with original exception and informative message in the parameters:
```
catch (Exception e) {
    throw new RuntimeException("Can't read data from the file " + fileName, e);
}
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,38 @@
...
+                BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName))) {
+            String values = reader.readLine();
+            int totalSupply = 0;
+            int totalBuy = 0;
+
+            while (values != null) {
+                String[] value = values.split(",");
+
+                if (value[0].equals("supply")) {
+                    totalSupply += Integer.parseInt(value[1]);
```
d1sam, line 21: `Integer.parseInt(value[1]);` you can extract this operation from if-else and here only use extracted variable.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,38 @@
...
+                } else {
+                    totalBuy += Integer.parseInt(value[1]);
```
d1sam, line 23: `Integer.parseInt(value[1]);` you can extract this operation from if-else and here only use extracted variable.

d1sam (changes requested):
Good job, but there are some mistakes need to be resolved.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,87 @@
...
+            throw new RuntimeException("Something went wrong "
+                    + fromFileName + " or " + toFileName, e);
+        }
+    }
+
+    private void createReport(BufferedWriter writer,
+                              int totalSupply, int totalBuy, int result) throws IOException {
+        writer.write("supply," + totalSupply + System.lineSeparator()
+                + "buy," + totalBuy + System.lineSeparator()
+                + "result," + result);
```
kshuryhin, line 29: Try to avoid magic-numbers and `String` literals, make them as a constants.

kshuryhin (changes requested):
Good job, but please, do not forget to resolve previous comments, especially related to String literals and constants across the code

kshuryhin (approved):
Good job!

PR #1163, opened 2023-10-01
added solution jv-work-with-file

TarasYurkovskyi:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+        return operationTotals;
+    }
 
+    private void writeReport(String fileName, int supplyTotal, int buyTotal, int result) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(SUPPLY + COMMA + supplyTotal);
+            writer.newLine();
+            writer.write(BUY + COMMA + buyTotal);
+            writer.newLine();
+            writer.write(RESULT + COMMA + result);
```
d1sam, line 67: It's better to use StringBuilder .append(), than String concatenation.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+                        System.err.println("Error parsing amount in line: " + line);
+                        e.printStackTrace();
+                    }
+                } else {
+                    System.err.println("Invalid line format: " + line);
+                }
+            }
+        } catch (IOException e) {
+            System.err.println("Can't read data from " + fileName);
+            e.printStackTrace();
```
d1sam, line 56: Better re-throw RuntimeException with original exception

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        try {
+            Map<String, Integer> operationTotals = readData(fromFileName);
+            int supplyTotal = operationTotals.getOrDefault(SUPPLY, DEFAULT_AMOUNT);
+            int buyTotal = operationTotals.getOrDefault(BUY, DEFAULT_AMOUNT);
+            int result = supplyTotal - buyTotal;
+            writeReport(toFileName, supplyTotal, buyTotal, result);
+        } catch (Exception e) {
+            System.err.println("Error processing files: " + e.getMessage());
+            e.printStackTrace();
```
d1sam, line 30: Leaving empty catch block or e.printStackTrace here is a bad practice. Better re-throw RuntimeException with original exception and informative message in the parameters:

```
catch (Exception e) {
    throw new RuntimeException("Can't read data from the file " + fileName, e);
}
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+    private void writeReport(String fileName, int supplyTotal, int buyTotal, int result) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(SUPPLY + COMMA + supplyTotal);
+            writer.newLine();
+            writer.write(BUY + COMMA + buyTotal);
+            writer.newLine();
+            writer.write(RESULT + COMMA + result);
+        } catch (IOException e) {
+            System.err.println("Can't write data to " + fileName);
+            e.printStackTrace();
```
d1sam, line 70: Leaving empty catch block or e.printStackTrace here is a bad practice. Better re-throw RuntimeException with original exception and informative message in the parameters:

catch (Exception e) {
    throw new RuntimeException("Can't read data from the file " + fileName, e);
}

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+                String[] parts = line.split(COMMA);
+                if (parts.length == REQUIRED_PARTS_COUNT) {
+                    String operation = parts[OPERATION_INDEX].trim();
+                    try {
+                        int amount = Integer.parseInt(parts[AMOUNT_INDEX].trim());
+                        operationTotals.put(operation, operationTotals
+                                .getOrDefault(operation, DEFAULT_AMOUNT) + amount);
+                    } catch (NumberFormatException e) {
+                        System.err.println("Error parsing amount in line: " + line);
+                        e.printStackTrace();
```
d1sam, line 48: Leaving empty catch block or e.printStackTrace here is a bad practice. Better re-throw RuntimeException with original exception and informative message in the parameters:

```
catch (Exception e) {
    throw new RuntimeException("Can't read data from the file " + fileName, e);
}

```

d1sam (changes requested):
Good job, but there are some mistakes need to be resolved.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        try {
+            Map<String, Integer> operationTotals = readData(fromFileName);
+            int supplyTotal = operationTotals.getOrDefault(SUPPLY, DEFAULT_AMOUNT);
+            int buyTotal = operationTotals.getOrDefault(BUY, DEFAULT_AMOUNT);
+            int result = supplyTotal - buyTotal;
+            writeReport(toFileName, supplyTotal, buyTotal, result);
+        } catch (Exception e) {
+            throw new RuntimeException("Error processing files " + fromFileName + toFileName, e);
+        }
```
mnyshenko, line 29: none of these methods throw any checked exception so there is no point in using try-catch here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+                String[] parts = line.split(COMMA);
+                if (parts.length == REQUIRED_PARTS_COUNT) {
+                    String operation = parts[OPERATION_INDEX].trim();
+                    try {
+                        int amount = Integer.parseInt(parts[AMOUNT_INDEX].trim());
+                        operationTotals.put(operation, operationTotals
+                                .getOrDefault(operation, DEFAULT_AMOUNT) + amount);
+                    } catch (Exception e) {
+                        throw new RuntimeException("Error parsing amount in line: " + line, e);
+                    }
```
mnyshenko, line 46: Why did you choose to use try-catch here?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+                } else {
+                    System.err.println("Invalid line format: " + line);
+                }
+            }
+        } catch (Exception e) {
+            throw new RuntimeException("Can't read data from the file " + fileName, e);
+        }
```
mnyshenko, line 53: Could we use a lower-level exception here?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+
+    private void writeReport(String fileName, int supplyTotal, int buyTotal, int result) {
+        StringBuilder reportData = new StringBuilder();
+        reportData.append(SUPPLY).append(COMMA).append(supplyTotal).append(System.lineSeparator());
+        reportData.append(BUY).append(COMMA).append(buyTotal).append(System.lineSeparator());
+        reportData.append(RESULT).append(COMMA).append(result);
 
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(reportData.toString());
+        } catch (Exception e) {
```
mnyshenko, line 65: Same

mnyshenko (changes requested):
A few points to consider

src/main/java/core/basesyntax/WorkWithFile.java
TarasYurkovskyi: Because there can be a NumberFormatException. For example if amount in our csv file will be "a123". Moreover, after amount there can be a space or Byte Order Mark and this exception will occur)

TarasYurkovskyi: There is trim() but it only removes characters with codes from 0 to 32 so there is a possibilitty for getting an exception here

TarasYurkovskyi: > Why did you choose to use try-catch here?

Because there can be a NumberFormatException. For example if amount in our csv file will be "a123". Moreover, after amount there can be other char like BOM) There is trim() but it only removes characters with codes from 0 to 32 so there is a possibilitty for getting an exception here

TarasYurkovskyi: > Why did you choose to use try-catch here?

Myb i should use replaceAll against trim() to not create a try catch block

kshuryhin (approved):
Great job!

PR #1164, opened 2023-10-02
Realization of WorkWithFile class

forzen123212:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+        StringBuilder report = new StringBuilder();
+        String[] allData = data.split(LINE_SEPARATOR);
+        int totalSupply = 0;
+        int totalBuy = 0;
+        for (String currentData : allData) {
+            String[] operationData = currentData.split(DELIMETER_COMA);
+            String operationType = operationData[INDEX_OF_TYPE];
+            String operationValue = operationData[INDEX_OF_VALUE];
+            if (operationType.equals(SUPPLY_NAME)) {
+                totalSupply += Integer.parseInt(operationValue);
```
d1sam, line 45: Integer.parseInt(operationValue); you can extract this operation from if-else and here only use extracted variable.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+            } else {
+                totalBuy += Integer.parseInt(operationValue);
```
d1sam, line 47: Integer.parseInt(operationValue); you can extract this operation from if-else and here only use extracted variable.

d1sam (approved):
Good job, I approve your homework, but please look at comment below :)

PR #1165, opened 2023-10-03
Implementation of the task "Work witch csv file".

druhov:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+                            supplyTotal += amount;
+                        } else if ("buy".equals(operationType)) {
+                            buyTotal += amount;
+                        }
+                    }
+                }
+
+                int result = supplyTotal - buyTotal;
+
+                writer.write("supply," + supplyTotal + System.getProperty("line.separator"));
```
okuzan, line 34: use System.lineSeparator() instead

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+
+                int supplyTotal = 0;
+                int buyTotal = 0;
+                String line;
+
+                while ((line = reader.readLine()) != null) {
+                    String[] parts = line.split(",");
+                    if (parts.length == 2) {
+                        String operationType = parts[0];
+                        int amount = Integer.parseInt(parts[1]);
```
okuzan, line 22: make ","  0 1 constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
okuzan, line 10: It isn't good practice to make long methods, especially public ones.
Please split your solution to three private methods like readFile, generateReport and writeToFile.
Then in your public method make only calls to there 3 methods to make it short and clear

okuzan (changes requested):
Please divide your logic before we can proceed with the review 🙏

okuzan (changes requested):
Not fixed, for separation of concerns purpose, please make your public method look like this
image

druhov:
> Not fixed, for separation of concerns purpose, please make your public method look like this image

Can you explain why you need to do this, I don't understand?

Ivan95kos (approved):
Nice

PR #1167, opened 2023-10-03
Added solution for a jv-work-with-file

AgroFix:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+
+    public String getResult(String inputFromFile) {
+        int buy = 0;
+        int supply = 0;
+        String[] lines = inputFromFile.split(System.lineSeparator());
+        for (String line : lines) {
+            String[] words = line.split(COMMA_SEPARATOR);
+            if (words.length == 2) {
+                if (words[0].equals(SUPPLY)) {
+                    supply += Integer.parseInt(words[1]);
```
d1sam, line 47: `Integer.parseInt(words[1]);` you can extract this operation from if-else and here only use extracted variable.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+                } else if (words[0].equals(BUY)) {
+                    buy += Integer.parseInt(words[1]);
```
d1sam, line 49: `Integer.parseInt(words[1]);` you can extract this operation from if-else and here only use extracted variable.

d1sam (approved):
Good job, but please pay attention to my comment below.

PR #1168, opened 2023-10-03
added solution to work-with-file homework

vadymhrnk:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+        int totalSupply = 0;
+        int totalBuy = 0;
+
+        try {
+            List<String> strings = Files.readAllLines(file.toPath());
+
+            for (String str : strings) {
+                String[] arr = str.split(",");
+                if (arr[INDEX_OF_STRING].equals(SUPPLY_STRING)) {
+                    totalSupply += Integer.valueOf(arr[INDEX_OF_VALUE]);
```
d1sam, line 28: `Integer.parseInt(arr[INDEX_OF_VALUE]); `you can extract this operation from if-else and here only use extracted variable.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+                } else if (arr[INDEX_OF_STRING].equals(BUY_STRING)) {
+                    totalBuy += Integer.valueOf(arr[INDEX_OF_VALUE]);
```
d1sam, line 30: `Integer.parseInt(arr[INDEX_OF_VALUE]); `you can extract this operation from if-else and here only use extracted variable.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file" + fromFileName, e);
+        }
```
d1sam, line 35: Please extract this logic to a separate method, e.g. "calculate report"

d1sam (changes requested):
Good job, but please review my comments and resolve some mistakes.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+        String report = createReport(totalSupply, totalBuy);
+        writeToFile(toFileName, report);
+    }
+
+    private String createReport(int totalSupply, int totalBuy) {
+        return String.format("%s,%d%n%s,%d%n%s,%d",
+                SUPPLY_STRING, totalSupply,
+                BUY_STRING, totalBuy,
+                RESULT_STRING, (totalSupply - totalBuy));
+
```
mnyshenko, line 32: Please don't leave redundant empty lines. Please check everywhere

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+    private static final String BUY_STRING = "buy";
+    private static final String RESULT_STRING = "result";
+    private static final String COMMA = ",";
+    private static final int ARRAY_FIRST_ELEMENT = 0;
+    private static final int ARRAY_SECOND_ELEMENT = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        File file = new File(fromFileName);
+        int totalSupply = calculateReport(file, fromFileName)[ARRAY_FIRST_ELEMENT];
+        int totalBuy = calculateReport(file, fromFileName)[ARRAY_SECOND_ELEMENT];
```
mnyshenko, line 21: Is there any point in calling calculateReport twice here?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+        int totalSupply = 0;
+        int totalBuy = 0;
+
+        try {
+            List<String> strings = Files.readAllLines(file.toPath());
+
+            for (String str : strings) {
+                String[] arr = str.split(COMMA);
+                String stringFromFile = arr[ARRAY_FIRST_ELEMENT];
+                int valueFromFile = Integer.parseInt(arr[ARRAY_SECOND_ELEMENT]);
```
mnyshenko, line 53: Please try to choose more descriptive variable names. arr, str are too generic

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+
+                if (stringFromFile.equals(SUPPLY_STRING)) {
+                    totalSupply += valueFromFile;
+                } else if (stringFromFile.equals(BUY_STRING)) {
+                    totalBuy += valueFromFile;
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file" + fromFileName, e);
```
mnyshenko, line 62: Please try to wrap code with try-catch which actually can throw an IOException. It's 48th in your case

mnyshenko (changes requested):
A few points to consider

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,74 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final String SUPPLY_STRING = "supply";
+    private static final String BUY_STRING = "buy";
+    private static final String RESULT_STRING = "result";
+    private static final String COMMA = ",";
+    private static final int ARRAY_FIRST_ELEMENT = 0;
```
bhdnchui, line 15: Maybe we can find more better name? 
If it is an index of the operation name, then `OPERATION_TYPE_INDEX`? Or something like this?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,74 @@
...
+    private static final int ARRAY_SECOND_ELEMENT = 1;
```
bhdnchui, line 16: Maybe we can find more better name? 
If it is an index of the value, then `VALUE_INDEX`? Or something like this?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,74 @@
...
+    }
+
+    private int[] calculateReport(List<String> strings) {
+        int totalSupply = 0;
+        int totalBuy = 0;
+
+        for (String pair : strings) {
+            String[] csvArray = pair.split(COMMA);
+            String stringFromFile = csvArray[ARRAY_FIRST_ELEMENT];
+            int valueFromFile = Integer.parseInt(csvArray[ARRAY_SECOND_ELEMENT]);
```
bhdnchui, line 63: Just `value`?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,74 @@
...
+        return strings;
+    }
+
+    private int[] calculateReport(List<String> strings) {
+        int totalSupply = 0;
+        int totalBuy = 0;
+
+        for (String pair : strings) {
+            String[] csvArray = pair.split(COMMA);
+            String stringFromFile = csvArray[ARRAY_FIRST_ELEMENT];
```
bhdnchui, line 62: Rename to `operationName`?

bhdnchui (approved):
Looks good. But one minor comment.
It would be better to declare the methods of the lower level in the order of their call by the method of the upper level.
```
public void getStatistic() {
  getData();
  calculate();
  drawResult();
}

private getData(){}
private calculate(){}
private drawResult(){}
```

PR #1177, opened 2023-10-04
Hw jv work with file

AdamMudrak:


okuzan (changes requested):
Good job on separating logic using two private methods, but you need to divide reading and report generating too. All literals in the code should be constants. Supply and buy shouldn't be fields

AdamMudrak:
So as I understand, you want me to use readFromFileAndCount1 rather than readFromFileAndCount.(there are two methods showcased in the initial solution). If I understood you correctly, theeen here we go.

olekskov (approved):

PR #1200, opened 2023-10-17
my solution_work_with_file

OlegShvets1:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+
     public void getStatistic(String fromFileName, String toFileName) {
```
okuzan, line 13: this method should contain three calls of other methods: readFile() - returns content, generateReport() -(you pass contents here), writeToFile()-(you pass report and filename here ), this will be a good code decomposition.example:

okuzan (changes requested):
All literals should be constants, use try with resources, check common mistakes

olekskov (approved):

PR #1204, opened 2023-10-19
workWithFile

lutumReal:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,41 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Arrays;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
kozhukhovsky, line 11: It is necessary to divide this method into 3 methods.
It should consist only of calls to private methods, for example:

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,41 @@
...
+        String newline;
+        int supplyTotal = 0;
+        int buyTotal = 0;
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName));
+                BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+
+            while ((newline = bufferedReader.readLine()) != null) {
+                String[] parts = newline.split(",");
+                if (parts.length == 2) {
+                    String category = parts[0];
```
kozhukhovsky, line 21: move indexes to the constant

kozhukhovsky (changes requested):

PR #1209, opened 2023-10-23
Made read and write

Kozharin-Vasul:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+                String [] option = value.split(",");
+                if (option[FIRST].equals(SUPPLY)) {
+                    result[FIRST] += Integer.valueOf(option[SECOND]);
+                } else if (option[FIRST].equals(BUY)) {
+                    result[SECOND] += Integer.valueOf(option[SECOND]);
+                }
+                value = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
```
okuzan, line 38: always include maximum details to make debugging easier => here, add fromFileName

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+        String report = generateReport(data);
+        writeToFile(toFileName, report);
+    }
+
+    private int[] readFromFile(String fromFileName) {
+        int[] result = new int[] {FIRST,FIRST};
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
+            String value = reader.readLine();
+            while (value != null) {
+                String [] option = value.split(",");
```
okuzan, line 29:
```suggestion
                String[] option = value.split(DELIMITER);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+    private static final int SECOND = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int[] data = readFromFile(fromFileName);
+        String report = generateReport(data);
+        writeToFile(toFileName, report);
+    }
+
+    private int[] readFromFile(String fromFileName) {
+        int[] result = new int[] {FIRST,FIRST};
```
okuzan, line 25:
```suggestion
        int[] result = new int[] {FIRST, FIRST};
```
don't forget to auto format your files before committing - Ctrl Alt L for windows

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final String SEPARATOR = ",";
+    private static final int FIRST = 0;
+    private static final int SECOND = 1;
```
okuzan, line 16: Consider giving meaningful names to your constants.
Naming constants just by their value is a missed opportunity to make the code more self-explanatory.
The name of a constant should indicate its role or purpose within the application, providing context to other developers.

Instead of:
// private static final int ZERO = 0;
// private static final int FIVE = 5;
// private static final String DASH = "-";

Consider something like:
// private static final int AMOUNT_INDEX = 0;
// private static final int DEFAULT_VALUE = 5;
// private static final Strint DELIMITER = "-";

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,59 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int[] data = readFromFile(fromFileName);
+        String report = generateReport(data);
```
okuzan, line 20: It would be better to have the readFromFile method focus solely on reading raw content from the file, without performing any data transformations or calculations. Right now, the method reads, parses, and calculates some initial values, returning an array with two elements.

This conflates multiple responsibilities and makes the function less reusable and harder to test. It's recommended to let all the parsing and calculations happen in generateReport. This would also improve code readability and maintainability by clearly separating the logic.

okuzan (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
Kozharin-Vasul: I am absolutely agree with it 
but especially in this case i will transfer redundant data and have to made redundant function to operate this data

Ivan95kos: your `readFromFile` method not only reads the file, it also makes calculations that must be performed in this `generateReport` method, and this is already a violation of the development principles of single responsibility, so it is better to redo it as Oleh recommended

Ivan95kos (changes requested):

kozhukhovsky (approved):
Good job!

PR #1214, opened 2023-10-25
implemented getStatistic, readCsvFile, createReport, writeToFile methods

yreuta:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final int TRANSACTION_TYPE_INDEX = 0;
+    private static final int TRANSACTION_VALUE_INDEX = 1;
+    private static final int SUPPLY_AMOUNT_INDEX = 0;
+    private static final int BUY_AMOUNT_INDEX = 1;
+    private static final int RESULT_AMOUNT_INDEX = 2;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int[] transactionsValues = readCsvFile(fromFileName);
```
Sanderfuck, line 21: Why you get array of int from file? I think it will be better to get in String value

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+                .append(transactionsValues[RESULT_AMOUNT_INDEX]).toString();
+    }
 
+    private void writeToFile(String report, String toFileName) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(report);
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("Error while creating file " + toFileName, e);
+        } catch (IOException e) {
+            throw new RuntimeException("Error while writing to file " + toFileName,e);
```
Sanderfuck, line 66: You cache 2 different exceptions and have the same handling, catch just IOException and thats will be enough

Sanderfuck (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+                stringFileContent.append(line).append(System.lineSeparator());
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fromFileName, e);
+        }
+        return stringFileContent.toString();
+    }
+
+    private String createReport(String fileContent) {
+        String[] transactions = fileContent.split(System.lineSeparator());
```
Rommelua, line 37: Always try to use words from the business task. So, 'operation' is more suitable than 'transaction' here and for constants (TRANSACTION_TYPE_INDEX)
image

Rommelua (approved):

PR #1223, opened 2023-11-05
exercise is soluted.

serhiimakarchuk:
maybe a lot of hardcode..

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,43 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
-
+        int valueSupply = 0;
+        int valueBuy = 0;
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
+            StringBuilder stringBuilder = new StringBuilder();
+            String value = reader.readLine();
+            while (value != null) {
+                String[] split = value.split(",");
+                value = reader.readLine();
+                if (split[0].equals("supply")) {
+                    valueSupply += Integer.parseInt(split[1]);
+                } else if (split[0].equals("buy")) {
+                    valueBuy += Integer.parseInt(split[1]);
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
+        }
+        String supply = "supply," + String.valueOf(valueSupply);
+        String buy = "buy," + String.valueOf(valueBuy);
+        String result = "result," + String.valueOf(valueSupply - valueBuy);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,43 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
okuzan, line 10: To improve code structure and readability, consider breaking down the process into three distinct methods: readFile(), generateReport(), and writeToFile(). Each method will have a clear responsibility, enhancing maintainability. For instance:

1. Use readFile(fromFileName) to fetch and return the file's contents.
2. Pass the content to generateReport(data) to create the desired report.
3. Finally, call writeToFile(toFileName, report) to save the report.

Applying this structure to your current code would not only simplify the getStatistic method but also follow good coding practices. Here's a quick outline:

This approach will keep your code organized and make each part easier to test and debug.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+
+        int valueSupply = 0;
+        int valueBuy = 0;
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
+            String value = reader.readLine();
+            while (value != null) {
+                String[] split = value.split(SEPARATOR);
+                value = reader.readLine();
+                if (split[OPERATION_INDEX].equals(SUPPLY)) {
```
okuzan, line 33: image

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+    private static final int VALUE_INDEX = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] data = readFile(fromFileName);
+        String report = generateReport(data);
+        writeToFile(toFileName, report);
+    }
+
+    private String [] readFile(String fromFileName) {
+
```
okuzan, line 24:
```suggestion
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+    private static final int OPERATION_INDEX = 0;
+    private static final int VALUE_INDEX = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] data = readFile(fromFileName);
+        String report = generateReport(data);
+        writeToFile(toFileName, report);
+    }
+
+    private String [] readFile(String fromFileName) {
```
okuzan, line 23: readFile shouldn't be involved in parsing data, just read as is, this will eliminate the need for your custom format of a two-elements array, which is a fragile approach. Just let generateReport method deal with it and you won't need to pass it between the methods

okuzan (changes requested):

PR #1236, opened 2023-11-14
Implemented method getStatistic()

SSosedova:
Implemented method getStatistic() that creates report and writes to file.


okuzan (changes requested):
Good job on the first try!

kozhukhovsky (approved):

PR #1256, opened 2023-12-08
Hello, I made 2 variants of solution to this task (two different commits in the branch named hm_VdBondarev). If it is possible for you, can you compare them and give me feedback which one is better to use?

vdbond:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+        int benefit = supplySum - buySum;
+        writeToFile(toFileName, supplySum, buySum, benefit);
+    }
+
+    public int getSupplySum(String[] dividedByEmptySpace) {
+        int supplySum = 0;
+        for (String info : dividedByEmptySpace) {
+            String[] dividedByComma = info.split(",");
+            String firstElement = dividedByComma[0];
+            int amountOfMoney = Integer.parseInt(dividedByComma[1]);
```
okuzan, line 24: create constants: SEPARATOR, AMOUNT_INDEX, OPERATION_TYPE_INDEX. DividedByComma, firstElement - are some bad naming examples, don't tell us anything about what value they store

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        String fileGotRead = readFromFile(fromFileName);
+        String[] dividedByEmptySpace = fileGotRead.split(" ");
+        int supplySum = getSupplySum(dividedByEmptySpace);
+        int buySum = getBuySum(dividedByEmptySpace);
+        int benefit = supplySum - buySum;
```
okuzan, line 15: encapsulate this logic in `generateReport` method which would return a string of a report

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+    }
 
+    public void writeToFile(String toFileName, int supplySum, int buySum, int benefit) {
+        StringBuilder result = new StringBuilder("supply,").append(supplySum)
+                .append("\n").append("buy,").append(buySum)
+                .append("\n").append("result,").append(benefit);
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(result.toString());
+        } catch (IOException e) {
+            throw new RuntimeException("There is an exception at line 36 named ", e);
```
okuzan, line 66: 1. include filename in the message
2. don't include a specific line mention in the code, it may change - in fact, it seems like it's already outdated :)

okuzan (changes requested):

okuzan (commented):
Haven't noticed the commit message, the second attempt is better - we should delegate the logic off to dedicated private methods

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+            throw new RuntimeException("There is an exception in (WorkWithFile) file named ", e);
+        }
+        return fullFileText.toString();
+    }
 
+    private void writeToFile(String toFileName, String data) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(data);
+        } catch (IOException e) {
+            throw new RuntimeException("There is an exception in (WorkWithFile) file named ", e);
```
okuzan, line 60:
```suggestion
            throw new RuntimeException("Couldn't write to this file: " + toFileName, e);
```

okuzan (changes requested):
Good job

olekskov (approved):
Good job

PR #1260, opened 2023-12-16
Implemented 'getStatistic' method.

davydov-software:


okuzan (approved):
Good job, interesting approach 👏

PR #1278, opened 2024-01-09
added solution to the task

yanna-stepanova:


kozhukhovsky (changes requested):
Divide all logic into 3 methods:
```
```
The getStatistic() method should only contain calls to these methods.

olekskov (approved):
Good job

PR #1283, opened 2024-01-17
implemented getStatistic method

Oleksii-Naumenko:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,54 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SEPARATOR = ",";
+    private static final int WORD_INDEX = 0;
+    private static final int NUMBER_INDEX = 1;
+
     public void getStatistic(String fromFileName, String toFileName) {
```
kozhukhovsky, line 15: It is necessary to break the logic into 3 methods. The getStatistic() method should only consist of calling these methods:

kozhukhovsky (changes requested):

olekskov (approved):
Good job

PR #1284, opened 2024-01-17
Implemented getStatistic method to read, sort, calculate data, and wr…

vfilimo:
…ite to a new file

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,57 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,57 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+    private static final int OPTION = 0;
+    private static final int VALUE = 1;
 
+    public void getStatistic(String fromFileName, String toFileName) {
+        int buy = 0;
```
kozhukhovsky, line 15: It is necessary to break the logic into 3 methods. The getStatistic() method should only consist of calling these methods:

kozhukhovsky (changes requested):

olekskov (approved):

PR #1289, opened 2024-01-22
File

MarynaBodina:
yjuj

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,47 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName));
+                BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName, true))) {
 
+            int buyTotal = 0;
+            int sellTotal = 0;
+
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] fields = line.split(",");
+                if (fields.length == 2) {
+                    String operationType = fields[0].trim().toLowerCase();
+                    int amount = Integer.parseInt(fields[1].trim());
+
+                    if ("buy".equals(operationType)) {
+                        buyTotal += amount;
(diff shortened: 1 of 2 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,48 @@
...
+                        buyTotal += amount;
+                    } else if ("supply".equals(operationType)) {
+                        supplyTotal += amount;
+                    }
+                }
+            }
+
+            final int result = supplyTotal - buyTotal;
+
+            // Измененные строки записи в файл
```
kozhukhovsky, line 35:
```suggestion
```

src/test/java/core/basesyntax/WorkWithFileTest.java
```diff
@@ -1,11 +1,11 @@
...
-import org.junit.Test;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
+        import org.junit.After;
+        import org.junit.Assert;
+        import org.junit.Test;
+        import java.io.IOException;
+        import java.nio.file.Files;
+        import java.nio.file.Path;
```
kozhukhovsky, line 8:
```suggestion
import org.junit.After;
import org.junit.Assert;
import org.junit.Test;
import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Path;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,48 @@
...
+                BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName, false))) {
+
+            int buyTotal = 0;
+            int supplyTotal = 0;
+
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] fields = line.split(",");
+                if (fields.length == 2) {
+                    String operationType = fields[0].trim().toLowerCase();
```
kozhukhovsky, line 22: move indexes to the class constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,48 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
kozhukhovsky, line 11: The method must be divided into 3 logical methods. As a result, the getStatistic() method will contain only calls to these methods and will look something like:

kozhukhovsky (changes requested):

PR #1294, opened 2024-01-25
Added solution to task

VladSchweps:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,68 @@
...
+        int result;
+        StringBuilder report = new StringBuilder();
+        for (String line : dataLine) {
+            String[] text = line.split(DIVIDER);
+            if (text[OPERATION_NAME].equals(OPPERATION_SUPPLY)) {
+                supplyTotal += Integer.parseInt(text[OPERATION_VALUE]);
+            } else if (text[OPERATION_NAME].equals(OPPERATION_BUY)) {
+                buyTotal += Integer.parseInt(text[OPERATION_VALUE]);
+            }
+        }
```
olekskov, line 51:
```suggestion
        for (String line : dataLine) {
            String[] text = line.split(DIVIDER);
            if (OPPERATION_SUPPLY.equals(text[OPERATION_NAME])) {
                supplyTotal += parseInt(text[OPERATION_VALUE]);
            } else if (OPPERATION_BUY.equals(text[OPERATION_NAME])) {
                buyTotal += parseInt(text[OPERATION_VALUE]);
            }
        }
```

olekskov (approved):
Good job, approved.

Some a bit advanced hints for future:
1) If you have constant or enum value it's generally better to call equals on it instead of some other value. In such way we make possibility of NPE much lower
2) when it's obvious from which class static method is called you may add static import to call method without class mentioned

PR #1298, opened 2024-01-27
add solution

mikke966:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,55 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 public class WorkWithFile {
+
+    private static final Logger LOGGER = Logger.getLogger(WorkWithFile.class.getName());
+
     public void getStatistic(String fromFileName, String toFileName) {
```
Ivan95kos, line 15: this method should contain three calls of other methods: readFile() - returns content, generateReport() -(you pass contents here), writeToFile()-(you pass report and filename here ), this will be a good code decomposition. example:
```
```

All magic numbers, strings, and symbols in your code should be constants.

Ivan95kos (changes requested):
Please, fix all checks

PR #1300, opened 2024-01-27
main version

EduarfF:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+    public static void getStatistic(String fromFileName, String toFileName) {
```
karma-o, line 11: once again, let's not make methods static

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+
```
karma-o, line 12:
```suggestion
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+                new FileReader(fileRead))) {
+
+            StringBuilder stringBuilder = new StringBuilder();
+
+            // Read from file
+            String value = reader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(" ");
+                value = reader.readLine();
+            }
```
karma-o, line 28: we can close this try block after we have read all of the strings to keep the opened connection to the file as short as possible

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+    public static void getStatistic(String fromFileName, String toFileName) {
+
+        File fileRead = new File(fromFileName);
+
+        int sumSupply = 0;
+        int sumBuy = 0;
+
+        try (BufferedReader reader = new BufferedReader(
+                new FileReader(fileRead))) {
+
```
karma-o, line 20:
```suggestion
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+            String[] splitBySpace = stringBuilder.toString().split(" ");
+
+            for (int i = 0; i < splitBySpace.length; ++i) {
+                String[] splitNameSum = splitBySpace[i].toString().split(",");
+                if (splitNameSum[0].equals("supply")) {
+                    sumSupply += Integer.parseInt(splitNameSum[1]);
+                } else {
+                    sumBuy += Integer.parseInt(splitNameSum[1]);
+                }
+            }
```
karma-o, line 39: lots of magic values, make them constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+            String value = reader.readLine();
+            while (value != null) {
+                stringBuilder.append(value).append(" ");
+                value = reader.readLine();
+            }
+
+            String[] splitBySpace = stringBuilder.toString().split(" ");
+
+            for (int i = 0; i < splitBySpace.length; ++i) {
+                String[] splitNameSum = splitBySpace[i].toString().split(",");
```
karma-o, line 33:
```suggestion
                String[] splitNameSum = splitBySpace[i].split(",");
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+                if (splitNameSum[0].equals("supply")) {
+                    sumSupply += Integer.parseInt(splitNameSum[1]);
+                } else {
+                    sumBuy += Integer.parseInt(splitNameSum[1]);
+                }
+            }
+
+            String[] resultSupply = new String[]{"supply", String.valueOf(sumSupply)};
+            String[] resultBuy = new String[]{"buy", String.valueOf(sumBuy)};
+            String[] resultResult = new String[]{"result", String.valueOf(sumSupply - sumBuy)};
```
karma-o, line 43: why do we need 3 array vars? it's much easier to just use simple string concatenation, of string builder, or string format would actually be the perfect solution.
We also won't need to use `String.valueOf()` in that case

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,64 @@
...
+
+            // Write to file
+            File writeFile = new File(toFileName);
+            try (BufferedWriter bufferedWriter = new BufferedWriter(
```
karma-o, line 47: nested file connection is opened here, which we should avoid, check common mistakes for details

src/main/java/core/basesyntax/WorkWithFile.java
karma-o, line 11: split your logic into 3 private methods as it is described in the common mistakes:
* `String[] readFromFile(String fileName)`
* `String[] formReport(String[] data)`
* `writeToFile(String[] data, String fileName)`
feel free to use just string instead of array

karma-o (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+        return stringBuilder.toString().split(" ");
+    }
+
+    private String[] getSumOperationType(String fromFileName) {
+        int sumSupply = 0;
+        int sumBuy = 0;
+        String[] readFromFile = readFromFile(fromFileName);
+        for (String s : readFromFile) {
+            String[] splitNameSum = s.split(",");
+            if (splitNameSum[OPERATION_TYPE_INDEX].equals("supply")) {
```
Elena-Bruyako, line 39: use constant for supply, buy, result

PR #1301, opened 2024-01-29
Work with file

ghost:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
kozhukhovsky, line 12: In this task you need to divide the logic into 3 methods.
As a result, this method will only contain calls to 3 methods:

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,45 @@
...
+        Map<String, Integer> operationMap = new HashMap<>();
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] parts = line.split(",");
+                if (parts.length == 2) {
+                    String operationType = parts[0].trim();
```
kozhukhovsky, line 20: move indexes to the constant

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+        String data = readFromFile(fromFileName);
+        String report = createReport(data);
+        writeToFile(toFileName, report);
+    }
+
+    private String readFromFile(String fromFileName) {
+        StringBuilder dataBuilder = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
+            String line;
+            final int operation_type_index = 0;
```
kozhukhovsky, line 22: it is not constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+            final int amount_index = 1;
+
+            while ((line = reader.readLine()) != null) {
+                String[] parts = line.split(",");
```
kozhukhovsky, line 26: "," repeated several times, it’s better to also put it in a constant

src/main/java/core/basesyntax/WorkWithFile.java
kozhukhovsky: https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

kozhukhovsky (approved):

PR #1312, opened 2024-02-06
solve the problem of processing ".csv" files

esidarapp:


kozhukhovsky (changes requested):
This method contains 3 logical blocks:
1. Read from file
2. Data processing and report generation
3. Write the report to a file

Move these 3 blocks into separate methods, and then the method will consist of only calling 3 methods:

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+        return "supply," + supplyCount + System.lineSeparator()
+                + "buy," + buyCount + System.lineSeparator()
+                + "result," + (supplyCount - buyCount);
+    }
 
+    private void writeToFile(String report, String fileName) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(fileName))) {
+            bufferedWriter.write(report);
+        } catch (IOException e) {
+            throw new RuntimeException("Problems with file!");
```
okuzan, line 55: always include filename to message

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+
+    private String createReport(String data) {
+        int supplyCount = 0;
+        int buyCount = 0;
+        String [] lines = data.split(System.lineSeparator());
+        for (String line : lines) {
+            String [] input = line.split(",");
+            switch (input[0]) {
+                case "supply":
+                    supplyCount += Integer.parseInt(input[1]);
```
okuzan, line 37: please introduce OPERATION_INDEX, AMOUNT_INDEX constants instead of magic literals

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+        }
+        return stringBuilder.toString();
+    }
+
+    private String createReport(String data) {
+        int supplyCount = 0;
+        int buyCount = 0;
+        String [] lines = data.split(System.lineSeparator());
+        for (String line : lines) {
+            String [] input = line.split(",");
```
okuzan, line 34: SEPARATOR

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+
+    private String readFromFile(String fileName) {
+        String line;
+        StringBuilder stringBuilder = new StringBuilder();
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fileName))) {
+            while ((line = bufferedReader.readLine()) != null) {
+                stringBuilder.append(line).append(System.lineSeparator());
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("problems with file!", e);
```
okuzan, line 24: same

okuzan (changes requested):
Good job!

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String data = readFromFile(fromFileName);
+        String report = createReport(data);
+        writeToFile(report,toFileName);
+    }
+
+    private String readFromFile(String fileName) {
+        String line;
+        StringBuilder stringBuilder = new StringBuilder();
```
okuzan, line 22: bad name, don't name after data type, but the role it serves, data it holds
`fileContent` would be better

src/main/java/core/basesyntax/WorkWithFile.java
okuzan: not fixed

okuzan (changes requested):

okuzan (approved):
Well done!

PR #1319, opened 2024-02-12
create methods

AnMishkin:
create readFile,calculetedSupplyForReport,calculetedBuyForReport,writeToFile

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,75 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
 
+public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+
+    public void getStatistic(String fromFileName, String toFileName) {
+        int supply = 0;
+        int buy = 0;
+
+        String allTextFile = readFile(fromFileName);
+        supply = calculetedSupplyForReport(allTextFile);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int supply = 0;
+        int buy = 0;
+
+        String allTextFile = readFile(fromFileName);
+        supply = calculetedSupplyForReport(allTextFile);
+        buy = calculetedBuyForReport(allTextFile);
+        writeToFile(supply,buy,toFileName);
+    }
```
olekskov, line 21: With current approach you have to go through file content twice what is not optimal.
Please replace calculetedSupplyForReport and calculetedBuyForReport with single method that will take all information with one loop though file content.
Also it would be better to move there logic that forms report from writeToFile method.
Overall from public method it should be something like below:
```suggestion
    public void getStatistic(String fromFileName, String toFileName) {
        String fileContent = readFile(fromFileName);
        String report = generateReport(fileContent);
        writeToFile(report, toFileName);
    }
```

olekskov (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,68 @@
...
+        }
+        resultBuilder.append(intSupply).append(',').append(intBuy);
+        return resultBuilder.toString();
     }
+
+    private void writeToFile(String report, String toFileName) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            if (report != null) {
+                String[] reportData = report.split(",");
+                bufferedWriter.write(SUPPLY + "," + reportData[0]);
```
Vasul-Pankiv, line 54: move magic numbers to constants

Vasul-Pankiv (approved):

PR #1321, opened 2024-02-14
jv-work-with-file HW

Cliffcamer:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+            throw new RuntimeException("Error processing file", e);
+        }
+        supplyTotal = 0;
+        buyTotal = 0;
+    }
+
+    private void processFile(String line, int supplyTotal, int buyTotal) {
+        String[] lines = line.split(",");
+        if (lines.length == 2) {
+            String operationType = lines[0].trim();
```
nowickimj, line 43: indexes can be externalized to constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private int supplyTotal;
```
nowickimj, line 10: you don't need to store this data as fields

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+                    processFile(line, supplyTotal, buyTotal);
+                }
+            }
+            int result = supplyTotal - buyTotal;
+
+            writeReport(toFileName, supplyTotal, buyTotal, result);
+        } catch (IOException e) {
+            throw new RuntimeException("Error processing file", e);
+        }
+        supplyTotal = 0;
```
nowickimj, line 36: if you remove fields, you don't need to reset them on every call

nowickimj (changes requested):

nowickimj (approved):

PR #1328, opened 2024-02-18
All assumptions met, tests passed

TylTejlor:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+            writer.write("supply," + supplyTotal);
+            writer.newLine();
+            writer.write("buy," + buyTotal);
+            writer.newLine();
+            writer.write("result," + result);
 
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
     }
```
Rommelua, line 43:
```suggestion
    public void getStatistic(String fromFileName, String toFileName) {
        String dataFromFile = readFile(fromFileName);
        String report = createReport(dataFromFile);
        writeToFile(report, toFileName);
    }
```
image

Rommelua (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+    private String createReport(String dataFromFile) {
+        int supplyTotal = 0;
+        int buyTotal = 0;
+        StringBuilder reportBuilder = new StringBuilder();
+
+        String[] lines = dataFromFile.split(System.lineSeparator());
+        for (String line : lines) {
+            String[] parts = line.split(",");
+            if (parts.length == 2) {
+                String operationType = parts[0];
```
nowickimj, line 40: Indexes can be stored as constants with descriptive name.

nowickimj (changes requested):
One small change requested, besides that looks great!

src/main/java/core/basesyntax/WorkWithFile.java
TylTejlor: So i should add it like that?:

public class WorkWithFile {

rest of the code...

can't set those constants inside the method body, or just before it, im obligated to declare them in this place, right?

nowickimj (approved):

PR #1333, opened 2024-02-20
Implemented the whole task logic

ghost:


Elena-Bruyako (approved):
Good job! 
Minor comment, LinkedHashMap is heavy data structured, so developers need to understand for which purposes they are going to use it. It depends what operations will be frequently used (like get, insert or remove data). Sometimes writing simple code is better then using a lot of memory

PR #1341, opened 2024-02-20
Added implementation of a task jv-work-with-file. Several new methods…

nklimovych:
… have been created to improve code decomposition

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,175 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
 
+/**
+ * This class provides methods to read data from a file, process it, and write the
+ * results to another file.
+ */
+public class WorkWithFile {
+    private static final String READ_DATA_EXCEPTION = "Can't read data from the file ";
+    private static final String WRITE_DATA_EXCEPTION = "Can't write data to the file ";
+    private static final String SUPPLY = "supply";
+    private static final String RESULT = "result,";
+    private static final String COMMA = ",";
+    private static final String SPACE = " ";
+    private static final int INDEX_ONE = 1;
+    private static final int INDEX_ZERO = 0;
+
+    /**
+     * Reads data from the specified file, processes the statistics, and writes the
+     * result to another file.
+     *
+     * @param fromFileName The name of the file to read data from.
+     * @param toFileName   The name of the file to write the result to.
+     */
+    public void getStatistic(String fromFileName, String toFileName) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,175 @@
...
+        }
+    }
+
+    /**
+     * Processes the raw data array and returns an array containing statistics.
+     *
+     * @param array The raw data array to process.
+     * @return An array containing processed statistics.
+     */
+    private String[] prepareStatistic(String[] array) {
```
nklimovych, line 84: This method is really big, but after a few hours of frustrating, it was the only solution I found to make sure that "supply" will be always in the first row...

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,175 @@
...
+                    amountList[0] = amount;
+                } else {
+                    int emptyIndex = findMatch(operationList, null);
+                    operationList[emptyIndex] = operation;
+                    amountList[emptyIndex] = amount;
+                }
+            }
+        }
+        return concatArrays(operationList, amountList);
+    }
```
nklimovych, line 113: This method is really big, but after a few hours of frustrating, it was the only solution I found to make sure that "supply" will be always in the first row...

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,131 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * This class provides methods to read data from a file, process it, and write the
```
andrii-hoienko, line 12: Please remove comments, they are redundant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,131 @@
...
+ * results to another file.
+ */
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final String COMMA = ",";
+    private static final String SPACE = " ";
+    private static final int INDEX_ONE = 1;
```
andrii-hoienko, line 21: Bad name, it is actually the same as someArray[1], no additional context

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,131 @@
...
+        return result;
+    }
 
+    /**
+     * Counts the number of unique elements in the array.
+     *
+     * @param list The array to count unique elements from.
+     * @return The number of unique elements in the array.
+     */
+    private int countUnicElements(String[] list) {
```
andrii-hoienko, line 122: I think this method is redundant, you can collect repost results to String/StringBuffer, this will avoid using array String[] result in generateReport method

andrii-hoienko (changes requested):
Nice, but some improvements required

andrii-hoienko (approved):
Nice

PR #1343, opened 2024-02-21
First solution

Nikname2303:


andrii-hoienko (changes requested):
Please read common mistakes, don`t forget about constants, magic number ect

PR #1345, opened 2024-02-21
Loop for read

azzilian:
compare by making arrays and if equals count their numbers Loop for write

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,46 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
 
+public class WorkWithFile {
+
+    public void getStatistic(String fromFileName, String toFileName) {
+        int totalSupply = 0;
+        int totalBuy = 0;
+        int totalResult = 0;
+        File incomingFile = new File(fromFileName);
+        File outgoingFile = new File(toFileName);
+        try (BufferedReader reader = new BufferedReader(new FileReader(incomingFile))) {
+            String i = reader.readLine();
+            while (i != null) {
+                String[] array = i.split(",");
+                String item = array[0].trim();
+                int summ = Integer.parseInt(array[1].trim());
+                if (item.equals("supply")) {
+                    totalSupply += summ;
+                } else if (item.equals("buy")) {
+                    totalBuy += summ;
+                }
+                i = reader.readLine();
+            }
+            totalResult = totalSupply - totalBuy;
+        } catch (IOException b) {
+            throw new RuntimeException(b.getMessage());
+        }
+
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(outgoingFile))) {
+            writer.write("supply," + totalSupply);
+            writer.newLine();
+            writer.write("buy," + totalBuy);
+            writer.newLine();
+            writer.write("result," + totalResult);
+        } catch (IOException e) {
+            throw new RuntimeException("Cannot write to file" + e.getMessage());
+        }
     }
 }


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+
     public void getStatistic(String fromFileName, String toFileName) {
```
andrii-hoienko, line 12: Please read common mistakes, also don`t forget about magic numbers and string constants

andrii-hoienko (changes requested):
Please read common mistakes, also don`t forget about magic numbers and string constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,60 @@
...
+        String amount = createReport(fromFileName);
+        writeToFile(toFileName, amount);
+    }
 
+    public String createReport(String fromFileName) {
+        String operation = "";
+        String fileLine = "";
+        int totalSupply = 0;
+        int totalBuy = 0;
+        int totalResult = 0;
```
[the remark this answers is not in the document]
azzilian: but those are not constants  and if I will make theme class level - then all calculations goes wrong

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,60 @@
...
+                .append(totalResult)
+                .toString();
     }
+
+    public void writeToFile(String toFileName, String amount) {
+        String[] count = amount.split(",");
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName))) {
+            writer.write("supply," + count[0] + System.lineSeparator());
+            writer.write("buy," + count[1] + System.lineSeparator());
+            writer.write("result," + count[2]);
```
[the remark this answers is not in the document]
andrii-hoienko: Not resolved

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,60 @@
...
+        StringBuilder builder = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(incomingFile))) {
+            fileLine = reader.readLine();
+            while (fileLine != null) {
+                String[] separatedItems = fileLine.split(",");
+                operation = separatedItems[0].trim();
+                if (operation.equals("supply")) {
+                    totalSupply += Integer.parseInt(separatedItems[1]);
+                } else if (operation.equals("buy")) {
+                    totalBuy += Integer.parseInt(separatedItems[1]);
```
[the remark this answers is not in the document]
andrii-hoienko: Not resolved

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,78 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static String fileLine;
```
andrii-hoienko, line 11: It is redundant to add all these variables to the class-level, they should me method level. It is possible to transfer all necessary information from method to method

andrii-hoienko (changes requested):
Some significant refactoring required, please read comment more carefully

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,80 @@
...
+        String line = readFromFile(fromFileName);
+        String report = createReport(line);
+        writeToFile(toFileName, report);
+    }
+
+    public String readFromFile(String fromFileName) {
+        File incomingFile = new File(fromFileName);
+        String readString;
+        StringBuilder readerBuilder = new StringBuilder();
+        try (BufferedReader reader = new BufferedReader(new FileReader(incomingFile))) {
```
karma-o, line 24: ideally, we should keep the try section as short as possible, to isolate the potential error from the rest of the code

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,80 @@
...
+            String fileLine = reader.readLine();
+            while (fileLine != null) {
+                readerBuilder.append(fileLine).append(",");
```
karma-o, line 27: `,` should be a constant
Overall it looks wrong to mix up the data into a single string without any other separators, brings us more problems, when we could have kept the line separator instead and worked with the lines that we could then split by comma and assign clear operation index, amount index, etc. 
So note this for the future

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,80 @@
...
+            }
+            readString = readerBuilder.toString();
+            readerBuilder.setLength(0);
+            return readString;
+        } catch (IOException b) {
+            throw new RuntimeException("Cannot read from file" + b.getMessage());
+        }
+    }
+
+    public String createReport(String line) {
```
karma-o, line 38: lots of literals and static numbers that should be constants, e.g. `2` `1` ` ` `,` `result`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,80 @@
...
+                .append(totalSupply)
+                .append(System.lineSeparator())
+                .append(BUY_TEXT)
+                .append(",")
+                .append(totalBuy)
+                .append(System.lineSeparator())
+                .append("result,")
+                .append(totalAmount)
+                .toString();
+        reportBuilder.setLength(0);
```
karma-o, line 68: why do you set length to 0?

karma-o (approved):
Approved but please check the comments

src/main/java/core/basesyntax/WorkWithFile.java
azzilian: if we will call method second time it will use builder that already contains some data and just will append new data to old data - multiply result in the end

src/main/java/core/basesyntax/WorkWithFile.java
azzilian: oh ok, just a lot of different opinions I heard about this like if we use it one time or not often -> no need for constants etc

src/main/java/core/basesyntax/WorkWithFile.java
azzilian: will do, thx

PR #1348, opened 2024-02-21
feat : first-commit

rel1c-hub:
Tests and build are successful.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,57 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashMap;
 
+public class WorkWithFile {
+    private static final String SUPPLY_KEY = "supply";
+    private static final String BUY_KEY = "buy";
+    private static final String RESULT_KEY = "result";
+    private static final String CSV_SPLIT_KEY = ",";
+
+    public void getStatistic(String fromFileName, String toFileName) {
+        HashMap<String, Integer> hashMap = readAndProcessData(fromFileName);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,81 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file! " + fromFileName, e);
+        }
+        return stringBuilder.toString();
+    }
+
+    private HashMap<String, Integer> processData(String fromFileName) {
+        HashMap<String, Integer> operationTotals = new HashMap<>();
+        String fileContent = read(fromFileName);
```
karma-o, line 41: would be nice to split it even more and call `read()` outside of this method, so that `processData()` would receive only `fileContent` and worked with that only

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,81 @@
...
+        operationTotals.merge(operationType, tempAmount, Integer::sum);
+    }
+
+    private int calculateResult(int supplySum, int buySum) {
+        return supplySum - buySum;
+    }
+
+    private String generateReport(HashMap<String, Integer> operationTotals) {
+        int supplyTotal = operationTotals.getOrDefault(SUPPLY_KEY, 0);
+        int buyTotal = operationTotals.getOrDefault(BUY_KEY, 0);
```
karma-o, line 65: I do not really get why we need the default value here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,81 @@
...
+        String[] values = line.split(CSV_SPLIT_KEY);
+        String operationType = values[OPERATION_TYPE_INDEX];
+        int tempAmount = Integer.parseInt(values[VALUE_INDEX]);
+
+        operationTotals.merge(operationType, tempAmount, Integer::sum);
+    }
+
+    private int calculateResult(int supplySum, int buySum) {
+        return supplySum - buySum;
+    }
```
karma-o, line 61: wouldn't say that this much separation is nessesary

karma-o (approved):
Looks good enough! Just please check some of my comments too

PR #1350, opened 2024-02-23
Implement method getStatistic(String fromFileName, String toFileName)…

Ololenko:
… which will return some data from the market after the working day.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,75 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+
 public class WorkWithFile {
+    private static final int INDEX_DATA = 0;
+    private static final int INDEX_MONEY = 1;
+    private static final String COMMA = ",";
+    private static final String SEPARATOR = System.lineSeparator();
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String data = readFromFile(fromFileName);
+        String report = createReport(data);
+        writeToFile(report, toFileName);
 
     }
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+    public void writeToFile(String report, String toFileName) {
+        File writeFile = new File(toFileName);
+        try {
+            Files.write(writeFile.toPath(), report.getBytes());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file");
+        }
+    }
 }
+
```
okuzan, line 75: two empty lines, should always be one

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+                .append("result,")
+                .append(totalAmount).toString();
+    }
+
+    public void writeToFile(String report, String toFileName) {
+        File writeFile = new File(toFileName);
+        try {
+            Files.write(writeFile.toPath(), report.getBytes());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file");
```
okuzan, line 71: always include filename to exception messages related to files

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+        String data = readFromFile(fromFileName);
+        String report = createReport(data);
+        writeToFile(report, toFileName);
 
     }
+
+    public String readFromFile(String fromFileName) {
+        File file = new File(fromFileName);
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(file));
```
okuzan, line 27: use try with resources
```suggestion
        try {
            BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName));
```

okuzan (changes requested):
Good job, clean code!

PR #1353, opened 2024-02-27
Task compleated

apgroupua:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+
 public class WorkWithFile {
+    private static final String COMMA = ",";
+    private static final String SEPARATOR = System.lineSeparator();
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final int DATA = 0;
+    private static final int MONEY = 1;
```
okuzan, line 15:
```suggestion
    private static final int DATA_INDEX = 0;
    private static final int MONEY_INDEX = 1;
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+                .append("result,")
+                .append(totalAmount).toString();
+    }
 
+    public void writeToFile(String report, String toFileName) {
+        File writeFile = new File(toFileName);
+        try {
+            Files.write(writeFile.toPath(), report.getBytes());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write file");
```
okuzan, line 72: always include filename to file-related exception messages

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+            StringBuilder builder = new StringBuilder();
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                builder.append(value).append(System.lineSeparator());
+                value = bufferedReader.readLine();
+            }
+            String readd = builder.toString();
+            return readd;
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
```
okuzan, line 36: same

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        String data = readFromFile(fromFileName);
+        String report = createReport(data);
+        writeToFile(report, toFileName);
+    }
+
+    public String readFromFile(String fromFileName) {
+        File file = new File(fromFileName);
+        try {
+            BufferedReader bufferedReader = new BufferedReader(new FileReader(file));
```
okuzan, line 26: BufferedReader implements AutoCloseable

okuzan (changes requested):
Good job! Minor fixes suggested

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+        File file = new File(fromFileName);
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
+            StringBuilder builder = new StringBuilder();
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                builder.append(value).append(System.lineSeparator());
+                value = bufferedReader.readLine();
+            }
+            String readd = builder.toString();
+            return readd;
```
okuzan, line 34:
```suggestion
            return builder.toString();
```
In situations like this, there is no need to store the object in a local variable if it will be immediately used. Instead, you can pass/use the expression directly without assigning it to a variable
readd. builder - bad naming

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        String data = readFromFile(fromFileName);
+        String report = createReport(data);
+        writeToFile(report, toFileName);
+    }
+
+    public String readFromFile(String fromFileName) {
+        File file = new File(fromFileName);
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(file))) {
+            StringBuilder builder = new StringBuilder();
```
okuzan, line 27:
```suggestion
            StringBuilder fileContentBuilder = new StringBuilder();
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+            String value = bufferedReader.readLine();
+            while (value != null) {
+                builder.append(value).append(System.lineSeparator());
+                value = bufferedReader.readLine();
+            }
```
okuzan, line 32:
```suggestion
            String line;
            while ((line = bufferedReader.readLine()) != null) {
                builder.append(line).append(System.lineSeparator());
            }
```
value - imprecise naming in this case

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,75 @@
...
+            String[] splitFile = element.split(COMMA);
+            if (splitFile[DATA_INDEX].equals(SUPPLY)) {
+                amountSupply += Integer.parseInt(splitFile[MONEY_INDEX]);
+            }
+            if (splitFile[DATA_INDEX].equals(BUY)) {
+                amountBuy += Integer.parseInt(splitFile[MONEY_INDEX]);
+            }
+        }
+        int totalAmount = amountSupply - amountBuy;
+        return value.append(SUPPLY)
```
okuzan, line 55:
```suggestion
        int totalAmount = amountSupply - amountBuy;
        
        return value.append(SUPPLY)
```
a matter of personal preference, but before bulky return we usually add an empty line

okuzan (changes requested):
some other minor fixes

PR #1354, opened 2024-02-27
Implemented the logic, created 2 new methods - createReport and write…

bilberryyy:
…ToFile, was used bufferedReader and writer

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,44 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        int supplyAmount = 0;
+        int buyAmount = 0;
 
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            String line;
+            while ((line = bufferedReader.readLine()) != null) {
+                String[] parts = line.split(",");
+                if (line.startsWith("supply")) {
+                    supplyAmount += Integer.parseInt(parts[1]);
+                } else {
+                    buyAmount += Integer.parseInt(parts[1]);
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fromFileName, e);
+        }
+        writeToFile(toFileName, supplyAmount, buyAmount);
+    }
+
+    public String createReport(int supplyAmount, int buyAmount) {
+        int resultAmount = supplyAmount - buyAmount;
+        return "supply," + supplyAmount + System.lineSeparator()
+                + "buy," + buyAmount + System.lineSeparator()
+                + "result," + resultAmount;
+    }
+
+    public void writeToFile(String toFileName, int supplyAmount, int buyAmount) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(createReport(supplyAmount, buyAmount));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write data to file" + toFileName, e);
+        }
     }
 }


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+                    buyAmount += Integer.parseInt(parts[1]);
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fromFileName, e);
+        }
+        writeToFile(toFileName, supplyAmount, buyAmount);
+    }
+
+    public String createReport(int supplyAmount, int buyAmount) {
```
okuzan, line 30: details of implementations shouldn't be exposed, choose a different access modifier
```suggestion
    private String createReport(int supplyAmount, int buyAmount) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+        int resultAmount = supplyAmount - buyAmount;
+        return "supply," + supplyAmount + System.lineSeparator()
+                + "buy," + buyAmount + System.lineSeparator()
+                + "result," + resultAmount;
+    }
 
+    public void writeToFile(String toFileName, int supplyAmount, int buyAmount) {
```
okuzan, line 37: same

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+                String[] parts = line.split(",");
+                if (line.startsWith("supply")) {
+                    supplyAmount += Integer.parseInt(parts[1]);
+                } else {
+                    buyAmount += Integer.parseInt(parts[1]);
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fromFileName, e);
+        }
```
okuzan, line 26: encapsulate this logic into private methods - one to read raw data as is, and a second one to parse it, we should [separate our concerns](https://en.wikipedia.org/wiki/Separation_of_concerns#:~:text=In%20computer%20science%2C%20separation%20of,code%20of%20a%20computer%20program.)

here (since business logic isn't super trivail), public method should only contain a few calls to the private methods, that do the actual work, more or less like this

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+    public String createReport(int supplyAmount, int buyAmount) {
+        int resultAmount = supplyAmount - buyAmount;
+        return "supply," + supplyAmount + System.lineSeparator()
+                + "buy," + buyAmount + System.lineSeparator()
+                + "result," + resultAmount;
+    }
 
+    public void writeToFile(String toFileName, int supplyAmount, int buyAmount) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(createReport(supplyAmount, buyAmount));
```
okuzan, line 39: methods are tightly coupled - let's make it more flexible, `writeToFile` shouldn't be bothered with creating a report, just pass a string/list of strings that need to be printed as an arg

okuzan (changes requested):
Good job overall! Refactor your code according to the comments below 👇

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,57 @@
...
+        return amount;
+    }
+
+    private String generateReport(String rawData) {
+        int supplyAmount = parseAmount(rawData, "supply");
+        int buyAmount = parseAmount(rawData, "buy");
+        int resultAmount = supplyAmount - buyAmount;
+        return "supply," + supplyAmount + System.lineSeparator()
+                + "buy," + buyAmount + System.lineSeparator()
+                + "result," + resultAmount;
```
MentorMA, line 47: Save strings to cnstants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,57 @@
...
+        return rawData.toString();
+    }
+
+    private int parseAmount(String rawData, String operationType) {
+        int amount = 0;
+        String[] lines = rawData.split(System.lineSeparator());
+        for (String line : lines) {
+            String[] parts = line.split(",");
+            if (parts[0].equals(operationType)) {
+                amount += Integer.parseInt(parts[1]);
```
MentorMA, line 35: Magic numbers to constants

MentorMA (changes requested):

olekskov (approved):
Good job

PR #1362, opened 2024-03-13
MateAcademy/Task/jv-work-with-file - Bartosz Wójcik solution

banderan:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,57 @@
...
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        File fileFrom = new File(fromFileName);
+        int buy = 0;
+        int supply = 0;
+        int result;
+        StringBuilder builder = new StringBuilder();
+
+        try {
```
nowickimj, line 17: This method is pretty huge - you might want to split it into smaller sub-processes, for example file reading part could be externalized as a method accepting `fromFileName` and returning the read string.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,57 @@
...
+        for (int i = 0; i < inputArray.length; i++) {
+            if (inputArray[i].equals("buy")) {
+                buy += Integer.parseInt(inputArray[i + 1]);
+            } else if (inputArray[i].equals("supply")) {
+                supply += Integer.parseInt(inputArray[i + 1]);
+            }
+        }
+        result = supply - buy;
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("supply,").append(supply).append(System.lineSeparator())
```
nowickimj, line 40: no need to split line 39 and 40 - you can start appending immediately after creating string builder.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,57 @@
...
+                .append("buy,").append(buy).append(System.lineSeparator())
+                .append("result,").append(result);
+        File file = new File(toFileName);
+
+        try {
```
nowickimj, line 45: file writing part could be externalized to a separate method.

nowickimj (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] inputArray = readingFromFile(fromFileName).split(",");
+        writingInToNewFile(toFileName, inputArray);
+    }
+
+    private static String readingFromFile(String fromFileName) {
```
nowickimj, line 15: 1. Please explain why is this method static.
2. Please use proper naming convention for methods e.g. `readFromFile()`

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+        }
+        result = supply - buy;
+        StringBuilder stringBuilder = new StringBuilder().append("supply,").append(supply)
+                .append(System.lineSeparator()).append("buy,").append(buy)
+                .append(System.lineSeparator()).append("result,").append(result);
+        return stringBuilder.toString();
+    }
+
+    private static void writingInToNewFile(String toFileName, String[] inputArray) {
+
```
nowickimj, line 54: Unnecessary new line

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+        File file = new File(toFileName);
+
```
nowickimj, line 56: Unnecessary new line

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+            }
+        }
+        result = supply - buy;
+        StringBuilder stringBuilder = new StringBuilder().append("supply,").append(supply)
+                .append(System.lineSeparator()).append("buy,").append(buy)
+                .append(System.lineSeparator()).append("result,").append(result);
+        return stringBuilder.toString();
+    }
+
+    private static void writingInToNewFile(String toFileName, String[] inputArray) {
```
nowickimj, line 53: Same here - align to common naming convention

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,69 @@
...
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] inputArray = readingFromFile(fromFileName).split(",");
+        writingInToNewFile(toFileName, inputArray);
```
nowickimj, line 12: The name suggests that you're trying to write an inputArray to file, but you're actually doing some operations on this file and then write results. Don't apply side effects to your code.

nowickimj (changes requested):
Please refactor the code to match the following concept:
1. read input from file
2. calculate results
3. write results to file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+    private String readFromFile(String fromFileName) {
+        File fileFrom = new File(fromFileName);
+        StringBuilder builder = new StringBuilder();
+
+        try {
+            BufferedReader reader = new BufferedReader(new FileReader(fileFrom));
+            String value = reader.readLine();
+            while (value != null) {
+                builder.append(value).append(System.lineSeparator());
+                value = reader.readLine();
```
okuzan, line 41:
```suggestion
            String value;
            while ((value = reader.readLine()) != null) {
                builder.append(value).append(System.lineSeparator());
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+            } else if (inputArray[i].equals("supply")) {
+                supply += Integer.parseInt(inputArray[i + 1]);
+            }
+        }
+        result = supply - buy;
+        StringBuilder stringBuilder = new StringBuilder().append("supply,").append(supply)
+                .append(System.lineSeparator()).append("buy,").append(buy)
+                .append(System.lineSeparator()).append("result,").append(result);
+        String calculateResult = stringBuilder.toString();
+
```
okuzan, line 28: extract this logic to a private method too, so your public method looks tidy and concise

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+        File file = new File(toFileName);
+
+        try {
+            file.createNewFile();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't create a file", e);
+        }
 
+        try {
+            Files.write(file.toPath(),calculateResult.getBytes());
```
okuzan, line 60:
```suggestion
            Files.write(file.toPath(), calculateResult.getBytes());
```
make sure to use IDEA's built-in formatter

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+        return builder.toString().replace(System.lineSeparator(), ",");
+    }
+
+    private void writeToNewFile(String toFileName, String calculateResult) {
+        File file = new File(toFileName);
+
+        try {
+            file.createNewFile();
+        } catch (IOException e) {
+            throw new RuntimeException("Can't create a file", e);
```
okuzan, line 56: it's a good idea to include filename to exceptions related to a file

okuzan (changes requested):
Good job, refactor the code so your method contains just three private method calls

src/main/java/core/basesyntax/WorkWithFile.java
nowickimj: Missed - please apply suggested changes.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Files;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] inputArray = readFromFile(fromFileName).split(",");
+        String calculateResult = editFile(inputArray);
```
nowickimj, line 12: This method name is not very helpful
1. First of all, you're not really editing any files - you read and process its content.
2. Name is too implicit - doesn't really tell anything what happens inside. "processTransaction()" or "calculateTransactionResults()" would make more sense.

src/main/java/core/basesyntax/WorkWithFile.java
nowickimj: Fixed.

nowickimj (changes requested):

nowickimj (approved):

PR #1367, opened 2024-03-17
First solution, basic tasks solved

noiseintheattic:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,55 @@
...
+                + buy + "," + buySum + System.lineSeparator()
+                + "result" + ","
+                + (supplySum - buySum)
+                + System.lineSeparator();
+    }
+
+    private void writeToFile(String report, String toFileName) {
+        File file = new File(toFileName);
 
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(file, false))) {
```
okuzan, line 49:
```suggestion
        try (BufferedWriter writer = new BufferedWriter(new FileWriter(file))) {
```
false is default

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,55 @@
...
+
+    private String readFromFile(String fileName) {
+        try {
+            return Files.readString(Path.of(fileName));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file " + fileName, e);
+        }
+    }
+
+    private String createReport(String dataFromFile) {
```
okuzan, line 23: I agree that createReport should accept dataFromFile, but you're actually passing filename, and calling read method inside this one. Let's separate our concerns and make methods more independent and flexible

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,55 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        writeToFile(createReport(fromFileName), toFileName);
```
okuzan, line 12:
```suggestion
        writeToFile(createReport(readFromFile(fromFileName)), toFileName);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,55 @@
...
+        }
+    }
+
+    private String createReport(String dataFromFile) {
+        final String supply = "supply";
+        final String buy = "buy";
+        int supplySum = 0;
+        int buySum = 0;
+
+        String[] arrayFromFile = readFromFile(dataFromFile).split("\\W+");
```
okuzan, line 29:
```suggestion
        String[] arrayFromFile = dataFromFile.split("\\W+");
```
could also extract regex to a constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,55 @@
...
+                + System.lineSeparator();
+    }
+
+    private void writeToFile(String report, String toFileName) {
+        File file = new File(toFileName);
 
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(file, false))) {
+            writer.write(report);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
okuzan, line 52: do the same as in read method:
```suggestion
            throw new RuntimeException("Can't write to the file " + fileName, e);```

okuzan (changes requested):
Good job!

src/main/java/core/basesyntax/WorkWithFile.java
noiseintheattic: ok, done

src/main/java/core/basesyntax/WorkWithFile.java
noiseintheattic: I hope I understood it correctly. done (I hope),

src/main/java/core/basesyntax/WorkWithFile.java
noiseintheattic: ok, done

src/main/java/core/basesyntax/WorkWithFile.java
noiseintheattic: ok, done

okuzan (approved):
Well done 👏

PR #1368, opened 2024-03-18
Update WorkWithFile.java

T0sh1s:


olekskov (commented):
https://github.com/mate-academy/jv-work-with-file/blob/master/checklist.md

Please recheck checklist items starting from

**Divide program logic
Large methods are very difficult to read and understand. So methods should be short, between 5-15 lines. And it's important for you to remember that one method should be responsible for one task.

For example: "write to file" or "create report" - are separate operations. So we should create separate methods for them.**

olekskov (changes requested):
Please divide logic and recheck other checklist items

PR #1380, opened 2024-03-29
Implemented method getStatistic

artyomwhite:


vadiimvooo (changes requested):
Sry, your task is appeard in FE mentors review. I will reject it to continue, just rerequest the review.

PR #1392, opened 2024-04-11
Initial solution

MarcusPy:
Few remarks:

- I used HashMap for this because I felt like it's a perfect opportunity to utilize them and gain some experience in them, on top of that, I believe this approach is pretty readable. Since they work very similarly to Python's dictionaries, it all felt very intuitive to me. 
- In order to satisfy the requirements placed upon by the automatic tests, I had to invert the order of Keys of the HashMap in line 32 and then introduce a backwards fori loop in the writeStatistic() method.
- I know I could have simply inserted "supply", "buy", "result" manually right after declaring the HashMap, but I really wanted to make my program capable of doing that dynamically if needed, as visible at line 17. This way it's more failproof and could archive additional data on the go, provided there were any.
- At line 28 I call a method and pass the second parameter in the form of another method. I was unable to find whether this is a bad practice or not. I would appreciate if You could clarify this for me.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,50 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashMap;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        HashMap<String, Integer> data = new HashMap<>();
+        try (BufferedReader br = new BufferedReader(new FileReader(fromFileName))) {
+            String value = br.readLine();
+            while (value != null) {
+                String[] str = value.split(",");
+                if (!data.containsKey(str[0])) {
+                    data.put(str[0], Integer.parseInt(str[1]));
+                } else {
+                    data.put(str[0], data.get(str[0]) + Integer.parseInt(str[1]));
+                }
+                value = br.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file", e);
+        }
+        data.put("result", data.get("supply") - data.get("buy"));
+        writeStatistic(toFileName, createStatistic(data));
+    }
 
+    private String createStatistic(HashMap<String, Integer> data) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,50 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashMap;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        HashMap<String, Integer> data = new HashMap<>();
```
nowickimj, line 12: The expected solution for this exercise is simpler than you might expect, based on your comment.

First things first: Please read the file as-is into string, no pre-processing is needed here.

I like the way that you split the process into separate steps. You can make three private methods:
```java
String readFile(fromFileName);
String createReport(data);
void writeFile(report, toFileName);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,50 @@
...
+        try (BufferedReader br = new BufferedReader(new FileReader(fromFileName))) {
+            String value = br.readLine();
+            while (value != null) {
+                String[] str = value.split(",");
+                if (!data.containsKey(str[0])) {
+                    data.put(str[0], Integer.parseInt(str[1]));
```
nowickimj, line 18: please make the indexes of values constants.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,50 @@
...
+                value = br.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file", e);
+        }
+        data.put("result", data.get("supply") - data.get("buy"));
+        writeStatistic(toFileName, createStatistic(data));
+    }
+
+    private String createStatistic(HashMap<String, Integer> data) {
```
nowickimj, line 31: This method should accept the string with content of a file, then:
1. Split entries by line separator
2. Split each line into values
3. Parse values and process operation
4. Return results

nowickimj (changes requested):

nowickimj (approved):
Awesome work! B-)

PR #1397, opened 2024-04-16
read, calculate, write with csv

BelitskyiAlexandr:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+            switch (parsingLine[OPERATION_TYPE_INDEX]) {
+                case "supply" :
+                    supply += amount;
+                    break;
+                case "buy" :
+                    buy += amount;
+                    break;
+                default:
+                    throw new RuntimeException("This type of operation is not supported");
+            }
```
Vasul-Pankiv, line 34: extract it into separate method too and make splitters and operationTypes a constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,73 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("The specified file does not exist");
+        }
+    }
 
+    private String createReport(int supply, int buy) {
+        StringBuilder sb = new StringBuilder();
+        sb.append("supply").append(",").append(supply).append(System.lineSeparator())
+                .append("buy").append(",").append(buy).append(System.lineSeparator())
+                .append("result").append(",").append(supply - buy)
```
Vasul-Pankiv, line 69: reuse constant values here

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
BelitskyiAlexandr: then we have to return two values from the method(buy, supply), and if we make them class fields, then we will have to create a new class object for each operation

kozhukhovsky (changes requested):
Separate the logic. The getStatistic method should consist of calling three methods:
```
```

BelitskyiAlexandr:
> Separate the logic. The getStatistic method should consist of calling three methods:
> 
> ```
> String dataFromFile = readFile(fromFileName);
> String report = createReport(dataFromFile);
> writeReportToFile(report, toFileName);
> ```

I do not understand your comment. I have split the logic into the three methods you described, in the method itself only the calculation takes place - what's wrong with that?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+
     public void getStatistic(String fromFileName, String toFileName) {
+        // required variables - are not class variables, because
+        // they do not reflect the state of the class
+        int supply = 0;
+        int buy = 0;
+
+        String[] lines = readFile(fromFileName).split(System.lineSeparator());
+
+        // calculating
```
MentorMA, line 25: comments in code not good practice of clean code

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+                    break;
+                case BUY:
+                    buy += amount;
+                    break;
+                default:
+                    throw new RuntimeException("This type of operation is not supported");
+            }
+        }
+
+        this.writeFile(toFileName, this.createReport(supply, buy));
```
MentorMA, line 41: image
Try to use this template for make good structure

MentorMA: parsing and calculation you can do in create report or add one more private method is it needed, up2u

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+    private void writeFile(String toFileName, String data) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(data);
+        } catch (IOException e) {
+            throw new RuntimeException("The specified file does not exist");
+        }
+    }
 
+    private String createReport(int supply, int buy) {
+        StringBuilder sb = new StringBuilder();
```
MentorMA, line 70: use full naming, `sb` - is bad pactice

MentorMA (changes requested):

sarakhmen (approved):
lgtm!

PR #1401, opened 2024-04-19
jv-work-with-file task solution

SvNadorozhnyi:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+
+    public void createReport(int supply, int buy, StringBuilder report) {
+        int result = supply - buy;
+        report.append("supply,")
+                .append(supply)
+                .append(System.lineSeparator())
+                .append("buy,")
+                .append(buy)
+                .append(System.lineSeparator())
+                .append("result,")
```
Elena-Bruyako, line 47: Make supply, buy, result and comma constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        StringBuilder report = new StringBuilder();
```
Elena-Bruyako, line 11: There is should be 3 methods - readFile, createReport and writeToFile.

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+    private static final String COMMA_SYMBOL = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        StringBuilder report = new StringBuilder();
+
+        readCalculateFile(fromFileName, report);
+        writeFile(toFileName, report);
+    }
+
+    public void readCalculateFile(String fromFileName, StringBuilder report) {
```
Elena-Bruyako, line 22:
```suggestion
    private void readCalculateFile(String fromFileName, StringBuilder report) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+                }
+                currentString = bufferedReader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from file " + fromFileName, e);
+        }
+        createReport(supply, buy, report);
+    }
+
+    public void createReport(int supply, int buy, StringBuilder report) {
```
Elena-Bruyako, line 44:
```suggestion
    private void createReport(int supply, int buy, StringBuilder report) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+                .append(supply)
+                .append(System.lineSeparator())
+                .append(STR_BUY)
+                .append(buy)
+                .append(System.lineSeparator())
+                .append(STR_RESULT)
+                .append(result);
+    }
 
+    public void writeFile(String toFileName, StringBuilder report) {
```
Elena-Bruyako, line 56:
```suggestion
    private void writeFile(String toFileName, StringBuilder report) {
```

src/main/java/core/basesyntax/WorkWithFile.java
Elena-Bruyako: First, don't need to declare variable that you don't actually use in this method 
Second, readCalculateFile has one task - read file, create report - is another task 
Look at common mistake file [Divide program logic](https://mate-academy.github.io/jv-program-common-mistakes/java-core/builder-file/work-with-file#divide-program-logic)

src/main/java/core/basesyntax/WorkWithFile.java
Elena-Bruyako, line 56: It's better to provide String report, because the logic of creating report should be in createReport() method

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String STR_SUPPLY = "supply,";
```
Elena-Bruyako, line 10: bad naming of constants. 
SUPPLY = "supply"; 
BUY = "buy";
Don't need to use comma in this constants, as you already have COMMA_SYMBOL because it gives flexibility to change value of constant when you need

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+    public void readCalculateFile(String fromFileName, StringBuilder report) {
+        int supply = 0;
+        int buy = 0;
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
+            String currentString = bufferedReader.readLine();
+            while (currentString != null) {
+                String[] splitedLine = currentString.split(COMMA_SYMBOL);
+                int operationAmount = Integer.parseInt(splitedLine[1]);
+                String operation = splitedLine[0];
+                if (operation.equals("supply")) {
```
Elena-Bruyako, line 31:
```suggestion
                if (operation.equals(SUPPLY)) {
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1414, opened 2024-05-08
Work with file homework solution

Experywow:
Added:
  readFromFile method:
    reading data from file, and return String[]
    with splitted data containing only information;
  makeReport method:
    Every call reset the StringBuilder object, for make sure we return renew actual data.
    Creating ArrayList for hold Operations such as supply,buy.
    Evaluate the amount for each operation type.
    Make report by StringBuilder object.
  writeToFile method:
    Record information stored in report to file.

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,74 @@
 package core.basesyntax;
 
-public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
 
+public class WorkWithFile {
+    private static final char CSV_DELIMITER = ',';
+    private static final String FILTER_FOR_DELIMITERS = "\\W+";
+    private static final String SUPPLY_FILTER = "supply";
+    private static final String BUY_FILTER = "buy";
+    private static final int RESET_STRING_BUILDER = 0;
+    private static final int INDEX_OF_FIRST_ELEMENT = 0;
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,74 @@
...
+        String report = makeReport(temperedData);
+        writeToFile(report, toFileName);
+    }
+
+    private String[] readFromFile(String fromFileName) {
+        String seizedData;
+        try {
+            seizedData = Files.readString(Path.of(fromFileName));
+        } catch (IOException e) {
+            throw new RuntimeException("Cant read from file..." + e);
```
sarakhmen, line 31: please specify the file name here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,74 @@
...
+        }
+        reportBuilder.append("result").append(CSV_DELIMITER).append(supplySum - buySum);
+        return reportBuilder.toString();
+    }
 
+    private void writeToFile(String report, String toFileName) {
+        try {
+            Files.write(Path.of(toFileName), report.getBytes());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to file..." + toFileName + " " + e);
```
sarakhmen, line 71: the same issue

sarakhmen (changes requested):
Good job, but let's slightly improve your solution :)

src/main/java/core/basesyntax/WorkWithFile.java
Experywow: You mean create uniqe name instead of reuse "fromFileName"?

Elena-Bruyako (approved):

PR #1427, opened 2024-05-29
Implement method to generate market statistics from CSV

smagles:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+            String value = reader.readLine();
+            while (value != null) {
+                builder.append(value).append(System.lineSeparator());
+                value = reader.readLine();
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(READ_ERROR_MESSAGE + fileName, e);
+        }
+        return builder.toString().split(System.lineSeparator());
+    }
```
MateuszRostek, line 58: Your solution is totally valid, see how you could do without using the `while` loop:) :

```suggestion
    private String[] readFromFile(String fromFileName) {
        File inputFile = new File(fromFileName);
        Path inputFilePath = Path.of(inputFile.toURI());
        List<String> lines;

        try {
            lines = Files.readAllLines(inputFilePath);
        } catch (IOException e) {
            throw new RuntimeException("Can't read from a file named: " + fromFileName, e);
        }
        return lines.toArray(new String[]{});
    }
```

MateuszRostek (approved):
Really Good Job, Approved! See minor advice.

PR #1434, opened 2024-06-08
hw-vdsfww work with file

vdsfww:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,68 @@
...
+                int amount = Integer.parseInt(parts[AMOUNT_INDEX]);
+
+                if (SUPPLY.equals(operationType)) {
+                    totalSupply += amount;
+                } else if (BUY.equals(operationType)) {
+                    totalBuy += amount;
+                }
+            }
+        }
+
```
okuzan, line 36: encapsulate into a separate method too ,so that your public method is clean and short

okuzan (changes requested):

sarakhmen (approved):
lgtm!

PR #1438, opened 2024-06-13
has implamented method that return some calculated data from given file

ArtemHurievskyi:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+    private static final int START_VALUE = 0;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        writeInfoToFile(calculateSupplyValue(readFromFile(fromFileName)),
+                calculateBuyValue(readFromFile(fromFileName)), toFileName);
+    }
+
+    private String[] readFromFile(String fromFileName) {
+        try (BufferedReader bufferedReader = new BufferedReader(
+                new FileReader(new File(fromFileName)))) {
```
sarakhmen, line 23: you can simply pass a filename instead of creating new File

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+            StringBuilder builder = new StringBuilder();
+            String line = bufferedReader.readLine();
+            while (line != null) {
+                builder.append(line).append(",");
```
sarakhmen, line 27: add a constant for this separator

sarakhmen: fix the same issue in all places

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+                line = bufferedReader.readLine();
+            }
+            return builder.toString().split(",");
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file" + e.getMessage());
```
sarakhmen, line 32: please specify the file name in your error message

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+
+    private void writeInfoToFile(int supply, int buy, String toFile) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(new File(toFile)))) {
+            bufferedWriter.write(SUPPLY + "," + supply + System.lineSeparator()
+                    + BUY + "," + buy + System.lineSeparator()
+                    + RESULT + "," + (supply - buy));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't create this file " + e.getMessage());
+        }
+    }
```
sarakhmen, line 64: same issues here

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+
```
sarakhmen, line 65:
```suggestion
```

sarakhmen (changes requested):
let's improve your solution :)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+                builder.append(line).append(COMA);
+                line = bufferedReader.readLine();
+            }
+            return builder.toString().split(COMA);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file " + fromFileName + e.getMessage());
+        }
+    }
+
+    private int calculateSupplyValue(String[] dataLines) {
```
Elena-Bruyako, line 36: It's better to create one method - calculateReport and pass it to writeIntoFile method
and return String as 
```
SUPPLY + "," + supply + System.lineSeparator()
                    + BUY + "," + buy + System.lineSeparator()
                    + RESULT + "," + (supply - buy)
```

but use StringBuilder to avoid creating new String every time during concatenation

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+                supplyValue += Integer.parseInt(dataLines[i + 1]);
+            } else if (dataLines[i].equals(BUY)) {
+                buyValue += Integer.parseInt(dataLines[i + 1]);
+            }
+        }
+        return new int[]{supplyValue, buyValue};
+    }
 
+    private void writeInfoToFile(int[] calculateReport, String toFile) {
+        StringBuilder builder = new StringBuilder();
```
okuzan, line 53: bad naming, because it doesn't tell us anything about what we expect to store in this variable - don't name List list, [] array etc, name is an opportunity to make code easier to read :)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFile))) {
+            builder.append(SUPPLY).append(COMMA).append(calculateReport[SUPPLY_INDEX])
+                    .append(System.lineSeparator()).append(BUY).append(COMMA)
+                    .append(calculateReport[BUY_INDEX])
+                    .append(System.lineSeparator()).append(RESULT).append(COMMA)
+                    .append(calculateReport[SUPPLY_INDEX] - calculateReport[BUY_INDEX]);
+            bufferedWriter.write(builder.toString());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write down to the file " + toFile + e.getMessage());
```
okuzan, line 62:
```suggestion
            throw new RuntimeException("Can't write down to the file " + toFile, e);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+                new FileReader(fromFileName))) {
+            StringBuilder builder = new StringBuilder();
+            String line = bufferedReader.readLine();
+            while (line != null) {
+                builder.append(line).append(COMMA);
+                line = bufferedReader.readLine();
+            }
+            return builder.toString().split(COMMA);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file " + fromFileName + e.getMessage());
```
okuzan, line 35: same

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        String[] separateDataFromFile = readFromFile(fromFileName);
+        int[] reportValues = calculateReport(separateDataFromFile);
+        writeInfoToFile(reportValues, toFileName);
+    }
+
+    private String[] readFromFile(String fromFileName) {
+        try (BufferedReader bufferedReader = new BufferedReader(
+                new FileReader(fromFileName))) {
+            StringBuilder builder = new StringBuilder();
```
okuzan, line 27: naming - same

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+            if (dataLines[i].equals(SUPPLY)) {
+                supplyValue += Integer.parseInt(dataLines[i + 1]);
+            } else if (dataLines[i].equals(BUY)) {
+                buyValue += Integer.parseInt(dataLines[i + 1]);
+            }
+        }
+        return new int[]{supplyValue, buyValue};
+    }
 
+    private void writeInfoToFile(int[] calculateReport, String toFile) {
```
okuzan, line 52: it's a bit fragile approach (returning array of int) - you're basically creating a custom format, and you need to uphold that contract in both methods, if any one of them changes, flow is broken. Not a big issue, but I would rather see the parsing logic in create report method, reader only reads given filename, writer only writes given string to given filename

okuzan (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,63 @@
...
+            return stringBuilder.toString().split(COMMA);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read from file " + fromFileName, e);
+        }
+    }
+
+    private String calculateReport(String[] dataLines) {
+        int supplyValue = START_VALUE;
+        int buyValue = START_VALUE;
+        StringBuilder stringBuilder = new StringBuilder();
```
okuzan, line 40:
```suggestion
        StringBuilder reportBuilder = new StringBuilder();
```

okuzan (approved):

PR #1442, opened 2024-06-20
finished hw

jakibczukw:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,35 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        File file = new File(fromFileName);
+        int supply = 0;
+        int buy = 0;
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            List<String> strings = Files.readAllLines(file.toPath());
 
+            for (String part : strings) {
+                String[] parts = part.split(",");
+                if (parts[0].equals("supply")) {
```
Vasul-Pankiv, line 20: make all hardcoded values constants. It is operation names, indexes, separators

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,39 @@
...
+import java.nio.file.Files;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final int OPERATION_PART = 0;
+    private static final int VALUE_PART = 1;
+    private static final int FIRST_VALUE = 0;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        File file = new File(fromFileName);
```
Elena-Bruyako, line 16: Create separate methods with these logic
read from file
generate report (calculcate report) 
write to file 


You don't need to create File file = new File(fromFileName); to implement the method

Elena-Bruyako (changes requested):

okuzan (approved):

PR #1448, opened 2024-06-29
.

vShmandiuk:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+            String value = reader.readLine();
+            while (value != null) {
+                builder.append(value).append(System.lineSeparator());
+                value = reader.readLine();
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(READ_ERROR_MESSAGE + fileName, e);
+        }
+        return builder.toString().split(System.lineSeparator());
+    }
```
MateuszRostek, line 58: Your solution is correct, here is an example of how it could be done using `Files` class:

```java
    private String[] readFromFile(String fromFileName) {
        Path inputFilePath = Path.of(inputFile.toURI());
        List<String> lines;
        try {
            lines = Files.readAllLines(inputFilePath);
        } catch (IOException e) {
            throw new RuntimeException("Can't read from a file named: " + fromFileName, e);
        }
        return lines.toArray(new String[]{});
    }
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+        int supplyAmount = 0;
+        int buyAmount = 0;
+        for (String entry : data) {
+            String[] parts = entry.split(DELIMITER);
+            String operationType = parts[OPERATION_TYPE_INDEX];
+            int amount = Integer.parseInt(parts[AMOUNT_INDEX]);
+            if (operationType.equals(SUPPLY)) {
+                supplyAmount += amount;
+            }
+            if (operationType.equals(BUY)) {
```
MateuszRostek, line 35:
```suggestion
            else if (operationType.equals(BUY)) {
```

MateuszRostek (approved):
Good job! See minor advice.

PR #1452, opened 2024-07-07
Implement method getStatistic(...);

aksoli666:
created classes Reader, Writer and
OperationService, ReportService

### DIFF
--- a/src/main/java/core/basesyntax/read/and/write/Reader.java
+++ b/src/main/java/core/basesyntax/read/and/write/Reader.java
@@ -1,0 +1,29 @@
+package core.basesyntax.read.and.write;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+
+public class Reader {
+    public String[] read(String fromFileName) {
+        try (BufferedReader reader = new BufferedReader(new FileReader(fromFileName))) {
+            String line;
+            String[] infoFromFile = new String[0];
+
+            while ((line = reader.readLine()) != null) {
+                infoFromFile = append(infoFromFile, line);
+            }
+
+            return infoFromFile;
+        } catch (IOException e) {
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/service/ReportService.java
```diff
@@ -0,0 +1,23 @@
...
+        int supply = salesVolume[SUPPLY_INDEX];
+        int buy = salesVolume[BUY_INDEX];
+        int result = salesVolume[RESULT_INDEX];
+
+        return new StringBuilder()
+                .append(SUPPLY).append(COMMA).append(supply).append(System.lineSeparator())
+                .append(BUY).append(COMMA).append(buy).append(System.lineSeparator())
+                .append(RESULT).append(COMMA).append(result).append(System.lineSeparator())
+                .toString();
+    }
```
MentorMateAcademy, line 22: better write in stack

src/main/java/core/basesyntax/read/and/write/Reader.java
```diff
@@ -0,0 +1,29 @@
...
+            throw new RuntimeException("Can`t read info from file: " + fromFileName, e);
+        }
+    }
+
+    private String[] append(String[] data, String line) {
+        String[] dataArray = new String[data.length + 1];
+        System.arraycopy(data, 0, dataArray, 0, data.length);
+        dataArray[data.length] = line;
+        return dataArray;
+    }
```
MentorMateAcademy, line 28: Why do you need this method, write it more simple, just read from file and save it it string format

MentorMateAcademy (changes requested):

src/main/java/core/basesyntax/read/and/write/Reader.java
aksoli666: For me more easier work with String array

Elena-Bruyako (approved):

PR #1464, opened 2024-07-21
commit task "work with file"

Oleksii21th:


MateuszRostek:
Please correct the code so that all tests pass. If you have a problem with fixing the code - go ahead and ask for help in the chat on our platform:)

MateuszRostek (changes requested):
.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,60 @@
...
+            bufferedWriter.newLine();
+            bufferedWriter.write("buy," + buy);
+            bufferedWriter.newLine();
+            bufferedWriter.write("result," + finalResult);
+        } catch (IOException ioException) {
+            throw new RuntimeException("Can't write data to the file " + toFileName, ioException);
+        }
+    }
+
+    private void readFromFileAndPutDataToMap(String fromFileName, Map<String, Integer> result) {
```
MateuszRostek, line 43: You should split this method into two separate. One for reading from the file, Second to create report (Or, as in your case, to change the read data from the file and write it to the map). Here is an example how read method should look:

image

MateuszRostek: image

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,60 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
 public class WorkWithFile {
```
MateuszRostek, line 12: Please use constants - using magic number/string literals is a bad approach.

```java
    private static final int OPERATION_VALUE_INDEX = 1;
```

MateuszRostek (changes requested):
Good job overall! There are some mistakes that should be corrected.

beatarjn (approved):

PR #1473, opened 2024-08-01
All tests passed. Created HashMap for statistics and writing them to new needed file

Wersebu:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
 public class WorkWithFile {
+
     public void getStatistic(String fromFileName, String toFileName) {
+        Map<String, Integer> statistics = readData(fromFileName);
+        writeData(toFileName, statistics);
```
MateuszRostek, line 15: image

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] parts = line.split(",");
+                if (parts.length != 2) {
+                    continue;
+                }
+
+                String operationType = parts[0].trim();
```
MateuszRostek, line 29: image

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+        writeData(toFileName, statistics);
+    }
+
+    private Map<String, Integer> readData(String fileName) {
+        Map<String, Integer> statistics = new HashMap<>();
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(fileName))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] parts = line.split(",");
```
MateuszRostek, line 24: image

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+                if (parts.length != 2) {
+                    continue;
+                }
+
+                String operationType = parts[0].trim();
+                int amount;
+                try {
+                    amount = Integer.parseInt(parts[1].trim());
```
MateuszRostek, line 32: image

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+import java.util.Map;
+
 public class WorkWithFile {
+
     public void getStatistic(String fromFileName, String toFileName) {
+        Map<String, Integer> statistics = readData(fromFileName);
+        writeData(toFileName, statistics);
+    }
+
+    private Map<String, Integer> readData(String fileName) {
```
MateuszRostek, line 18: You can use `java.nio.file.Files` to read all lines at once and then just return an array of String instead of using map. For example like this:

image

MateuszRostek (changes requested):
Before creating a PR, ALWAYS go through `common mistakes` provided at the end of the task description

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,65 @@
...
+    }
+
+    private String createReport(String[] lines) {
+        int supplyTotal = 0;
+        int buyTotal = 0;
+
+        for (String line : lines) {
+            String[] parts = line.split(",");
+            String operation = parts[0];
+            int amount = Integer.parseInt(parts[1]);
```
MateuszRostek, line 39: You still have magic numbers and string literals here, you should create constants and use them here, for example:

```java
public class WorkWithFile {
    private static final int OPERATION_VALUE_INDEX = 1;
```

And therefore you code should look like this:

```suggestion
            String[] parts = line.split(SEPARATOR);
            String operation = parts[OPERATION_TYPE_INDEX];
            int amount = Integer.parseInt(parts[OPERATION_VALUE_INDEX]);
```

MateuszRostek (approved):
Good job! There are some problems with the constants, but I'm sure you can handle it, approved!

PR #1482, opened 2024-08-09
implemented the csv processing capabilities

NikitaLafinskiy:


NikitaLafinskiy:
I considered implementing the StringBuilder towards the end of the method, however when I do it underlines it saying "StringBuilder could be replaced with a String" and prompts me to autocomplete the notice, when I do it just replaces the StringBuilder with a String, so as I understand it is not needed in this scenario

sarakhmen (changes requested):
Let's split your getStatistic method into 3 smaller ones: for read, process, write operations. Also fix magic numbers/literals

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+
+    private int[] processStatistics(String[] data) {
+        int supplyCount = STARTING_COUNT;
+        int buyCount = STARTING_COUNT;
+
+        for (String value : data) {
+            String[] currentLine = value.split(SEPARATOR);
+            if (currentLine.length == ELEMENTS_ON_THE_LINE) {
+                int count = Integer.parseInt(currentLine[1]);
+                if (currentLine[0].contains(SUPPLY)) {
```
okuzan, line 50: make indexes constants like OPERATION_INDEX

okuzan (changes requested):
one thing left!

okuzan (approved):

PR #1496, opened 2024-09-01
Solution

vsupruniuk:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
 package core.basesyntax;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.List;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
liliia-ponomarenko, line 9: Divide program logic
Large methods are very difficult to read and understand. So methods should be short, between 5-15 lines. And it's important for you to remember that one method should be responsible for one task.

For example: "write to file" or "create report" - are separate operations. So we should create separate methods for them.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+
+        File fromFile = new File(fromFileName);
+        File toFile = new File(toFileName);
+
+        List<String> fileData = null;
+
+        try {
+            fileData = Files.readAllLines(fromFile.toPath());
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
liliia-ponomarenko, line 21: Let's add informative message ;)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+        }
+
+        for (String data : fileData) {
+            String[] values = data.split(",");
```
liliia-ponomarenko, line 25: Make the delimiter a constant ;)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+                .append(buyAmount)
+                .append(System.lineSeparator())
+                .append("result,")
+                .append(result)
+                .append(System.lineSeparator());
 
+        try {
+            Files.write(toFile.toPath(), stringBuilder.toString().getBytes());
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
liliia-ponomarenko, line 55: the same

liliia-ponomarenko (approved):
Good job! Let's fix a few mistakes ;)

PR #1506, opened 2024-09-13
Implemented getStatistic method for processing CSV files

Petliuk:


okuzan (approved):
all literals better be constants

PR #1518, opened 2024-09-24
Solution

ostapzat:


Elena-Bruyako (changes requested):
refactor logic according to the comment

PR #1521, opened 2024-09-25
homework solution

sergiikonov:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,48 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String DELIMITER = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
```
Elena-Bruyako, line 12: Divide logic into separate methods
```
readFromFile
calculateResult
writeIntoFile
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+                int amount = Integer.parseInt(parts[1].trim());
+
+                if (operationType.equals(BUY)) {
+                    amountBuy += amount;
+                } else if (operationType.equals(SUPPLY)) {
+                    amountSupply += amount;
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file", e);
```
ahoienko, line 42:
```suggestion
            throw new RuntimeException("Can't read the file:" + fromFileName, e);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+            while ((line = reader.readLine()) != null) {
+                String[] parts = line.split(DELIMITER);
+                String operationType = parts[0].trim();
+                int amount = Integer.parseInt(parts[1].trim());
+
+                if (operationType.equals(BUY)) {
+                    amountBuy += amount;
+                } else if (operationType.equals(SUPPLY)) {
+                    amountSupply += amount;
+                }
```
ahoienko, line 39: This part should belong to method that makes calculations

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+        amountBuy = 0;
+        amountSupply = 0;
+        result = 0;
+
+        readFile(fromFileName);
+        calculateResult();
+        writeIntoFile(toFileName);
+    }
+
+    public void readFile(String fromFileName) {
```
ahoienko, line 27:
```suggestion
    public String readFile(String fromFileName) { //or List<String>
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read the file", e);
+        }
+    }
+
+    public void calculateResult() {
+        result = amountSupply - amountBuy;
+    }
```
ahoienko, line 48: This method should return something that will be provided to method writeIntoFile it could be a list with results (amountBuy, amountSupply, result) or array or map, or even custom result object

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
 
+    private void writeIntoFile(String toFileName) {
```
ahoienko, line 50:
```suggestion
    private void writeIntoFile(String toFileName, some data that should we written) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String DELIMITER = ",";
+    private static final String BUY = "buy";
+    private static final String SUPPLY = "supply";
+    private int amountBuy = 0;
+    private int amountSupply = 0;
+    private int result;
```
ahoienko, line 15:
```suggestion

```

ahoienko: It is bad to keep this data on class level. For example you created a object of WorkWithFile  class, to calculate some data, when you first calculated everything will be ok, but when you will try to calculate again using same object, this object will keep data about previous calculations, and It will not be possible to reuse it. You will need to create new object of WorkWithFile class to make new calculation. Or as another option you need to clear amountBuy, amountSupply and result after each calculation, but I don`t recommend this approach because it is more complicated and less clear then other options

ahoienko (changes requested):
Very nice) Lets improve your solution

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] parts = line.split(DELIMITER);
+                String operationType = parts[0].trim();
+                int amount = Integer.parseInt(parts[1].trim());
+
+                int[] updatedAmounts = processOperation(operationType,
+                        amount, amountBuy, amountSupply);
+                amountBuy = updatedAmounts[0];
+                amountSupply = updatedAmounts[1];
```
MentorMateAcademy, line 33: Magic numbers

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+        int result = amountSupply - amountBuy;
+        return new int[]{amountSupply, amountBuy, result};
+    }
 
+    private void writeIntoFile(String toFileName, int[] results) {
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write("supply," + results[0] + System.lineSeparator());
+            bufferedWriter.write("buy," + results[1] + System.lineSeparator());
+            bufferedWriter.write("result," + results[2]);
+        } catch (IOException e) {
```
MentorMateAcademy, line 62: write to file already prepared report

MentorMateAcademy (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,77 @@
...
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            bufferedWriter.write(report);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to the file: " + toFileName, e);
+        }
+    }
+
+    private String prepareReport(int[] results) {
+        StringBuilder stringBuilder = new StringBuilder();
+        stringBuilder.append("supply,").append(results[0]).append(System.lineSeparator());
```
Elena-Bruyako, line 71: use constants for supply, buy, result and comma
also use local variables for results[0], results[1] and results[2]

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1536, opened 2024-10-04
performed the logic of reading from the file and writing to the file,…

SwtLouisiana:
… also calculated the value

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -1,7 +1,60 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        List<Integer> supplyAmounts = new ArrayList<>();
+        List<Integer> buyAmounts = new ArrayList<>();
 
+        try (BufferedReader fromFile = new BufferedReader(new FileReader(fromFileName))) {
+            String line;
+
+            while ((line = fromFile.readLine()) != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,79 @@
...
+        writeResults(toFileName, statistics);
+    }
+
+    public void readAndParseData(String fromFileName, List<Integer> supplyAmounts,
+                                  List<Integer> buyAmounts) {
+        try (BufferedReader fromFile = new BufferedReader(new FileReader(fromFileName))) {
+            String line;
+
+            while ((line = fromFile.readLine()) != null) {
+                String[] values = line.split(",");
```
Elena-Bruyako, line 33: create constant for comme

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,79 @@
...
+
+                if (values.length == 2) {
+                    String operationType = values[0].trim();
```
Elena-Bruyako, line 36: create constants for 0 and 1

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,79 @@
...
+        int totalSupply = sumList(supplyAmounts);
+        int totalBuy = sumList(buyAmounts);
+        int result = totalSupply - totalBuy;
+
+        return new int[]{totalSupply, totalBuy, result};
+    }
 
+    public void writeResults(String toFileName, int[] statistics) {
+        try (BufferedWriter toFile = new BufferedWriter(new FileWriter(toFileName))) {
+            toFile.write(SUPPLY + "," + statistics[0]);
```
Elena-Bruyako, line 70: the logic of creating report should be in separate method
Also, use StringBuilder to concatenate strings

Elena-Bruyako (changes requested):

PR #1544, opened 2024-10-14
Hw file work

sannchous:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+    private static final String SUPPLY = "supply";
+    private static final String RESULT = "result";
+    private static final String BUY = "buy";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String text = readFileAndCreateReport(fromFileName);
+        writeReportToFile(toFileName, text);
+    }
 
+    private String readFileAndCreateReport(String fromFile) {
```
liliia-ponomarenko, line 21: When you use `And` in the method name it means that you should divide logic into separate methods (one method - one responsibility). Let's first read the file and pass the data to another method.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+        int supplyAmount = 0;
+        int buyAmount = 0;
+        int resultAmount;
+        StringBuilder reportBuilder = new StringBuilder();
+        try (BufferedReader reader
+                     = new BufferedReader(new FileReader(fromFile))) {
+            String data = reader.readLine();
+            while (data != null) {
+                String[] line = data.split(",");
```
liliia-ponomarenko, line 30: Make COMA constant field (you use it 4 times)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+                if (line[TYPE_INDEX].equals(SUPPLY)) {
+                    supplyAmount += Integer.parseInt(line[AMOUNT_INDEX]);
+                } else {
+                    buyAmount += Integer.parseInt(line[AMOUNT_INDEX]);
+                }
+                data = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file", e);
```
liliia-ponomarenko, line 39: Agree with AI-mentor. Let's pass the file name parameter to the exception message

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+                .append(BUY).append(",").append(buyAmount).append(System.lineSeparator())
+                .append(RESULT).append(",").append(resultAmount).toString();
+    }
+    
+    private void writeReportToFile(String toFile, String text) {
+        try (BufferedWriter writer
+                     = new BufferedWriter(new FileWriter(toFile))) {
+            writer.write(text);
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write file", e);
```
liliia-ponomarenko, line 53: the same

liliia-ponomarenko (changes requested):
Good job! Let's improve your solution ;)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+    private static final String RESULT = "result";
+    private static final String BUY = "buy";
+    private static final String COMMA = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
+        String text = createReport(fromFileName);
+        writeReportToFile(toFileName, text);
+    }
+
+    private int[] readFile(String fromFile) {
```
liliia-ponomarenko, line 22: in fact, this method not only reads data but also calculates it

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+        int supplyAmount = data[0];
+        int buyAmount = data[1];
+        int resultAmount = data[2];
+
+        StringBuilder reportBuilder = new StringBuilder();
+        return reportBuilder.append(SUPPLY).append(COMMA)
+                .append(supplyAmount).append(System.lineSeparator())
+                .append(BUY).append(COMMA).append(buyAmount).append(System.lineSeparator())
+                .append(RESULT).append(COMMA).append(resultAmount).toString();
+
```
liliia-ponomarenko, line 60:
```suggestion
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+                if (line[TYPE_INDEX].equals(SUPPLY)) {
+                    supplyAmount += Integer.parseInt(line[AMOUNT_INDEX]);
+                } else {
+                    buyAmount += Integer.parseInt(line[AMOUNT_INDEX]);
+                }
+                data = reader.readLine();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read file" + fromFile, e);
+        }
```
liliia-ponomarenko, line 41:
```suggestion
List<String> lines = new ArrayList();
        try (BufferedReader reader
                     = new BufferedReader(new FileReader(fromFile))) {
            String data = reader.readLine();
            while (data != null) {
                lines.add(data);
                data = reader.readLine();
            }
        } catch (IOException e) {
            throw new RuntimeException("Can't read file" + fromFile, e);
        }
        return lines;
```
Here reading logic. You can read data and pass it to another method that calculates data (createReport method)

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+        resultAmount = supplyAmount - buyAmount;
+        result[0] = supplyAmount;
+        result[1] = buyAmount;
+        result[2] = resultAmount;
+        return result;
+    }
+
+    private String createReport(String fromFile) {
+        int[] data = readFile(fromFile);
```
liliia-ponomarenko, line 50: it is better to call all private methods in a public method and pass the result as an input parameter to the next processing method

liliia-ponomarenko (changes requested):
Good job! ;) Let's refactor your solution ;)

PR #1545, opened 2024-10-17
hw

Anasteisha4:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,49 @@
 package core.basesyntax;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
ahoienko, line 10: This method is too complicated and it is difficult to read and understand. I suggest you to break it down into several methods. for example: readFile, calculateStatistics, generateReport and writeToFile. After that code will be much more readable, reusable and maintainable

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,49 @@
...
+        int totalSupply = 0;
+        int totalBuy = 0;
+
+        try (BufferedReader br = new BufferedReader(new FileReader(fromFileName))) {
+            String line;
+            while ((line = br.readLine()) != null) {
+                String[] parts = line.split(",");
```
ahoienko, line 17: Please make all string literals that you use constants

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,49 @@
...
+                if (parts.length != 2) {
+                    continue;
+                }
+                String operationType = parts[0].trim();
+                int amount;
+                try {
+                    amount = Integer.parseInt(parts[1].trim());
```
ahoienko, line 24: Please avoid using magic numbers, make them constants instead

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,49 @@
...
+        StringBuilder report = new StringBuilder();
+        int result = totalSupply - totalBuy;
+        report.append("supply,").append(totalSupply).append(System.lineSeparator());
+        report.append("buy,").append(totalBuy).append(System.lineSeparator());
+        report.append("result,").append(result).append(System.lineSeparator());
 
+        try (BufferedWriter wr = new BufferedWriter(new FileWriter(toFileName))) {
+            wr.write(report.toString());
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write file", e);
```
ahoienko, line 46:
```suggestion
            throw new RuntimeException("Can't write file: " + fileName, e);
```

ahoienko (changes requested):
Please read common mistakes carefully

PR #1570, opened 2024-11-27
Solution

MaybeTI:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,49 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String SUPPLY = "supply";
+    private static final String BUY = "buy";
+    private static final String RESULT = "result";
+    private static final String COMMA = ",";
+
     public void getStatistic(String fromFileName, String toFileName) {
```
liliia-ponomarenko, line 15: You need to split the logic into three private methods and call them in the public one by one.
1) Only reads data from the file and returns it.
2) Calculates the data and returns a report
3) Only writes the report to the file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,49 @@
...
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from file", e);
+        }
+        return total;
+    }
 
+    private void writeToFile(String toFileName, int supply, int buy) {
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName))) {
+            writer.append(SUPPLY).append(COMMA).append(String.valueOf(supply)).append("\n");
```
liliia-ponomarenko, line 42: Use System.lineseparator()

liliia-ponomarenko (changes requested):
Good job! Let’s  improve your solution ;)

PR #1572, opened 2024-11-27
Implement report statistic for market.

Kitazuka:


liliia-ponomarenko (changes requested):
You need to divide the logic into three methods and call them sequentially in a public method: 
1) only reads the file and returns a list of lines 
2) calculates data and returns a report 
3) only write the report to a file

PR #1574, opened 2024-12-01
add base solution

arastorhuiev:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,95 @@
...
 public class WorkWithFile {
+    private static final String STRING_SEPARATOR = ",";
+    private static final String[] REPORT_COLUMNS = new String[] {"supply", "buy", "result"};
+    private static BufferedWriter bufferedWriter = null;
+    private static File file = null;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        HashMap<String, Integer> results = calculateStatisticResults(fromFileName);
+        clearFile(toFileName);
+        file = new File(toFileName);
```
liliia-ponomarenko, line 20: You need to split the logic into three private methods and call them one by one in the public method.
1) only reads and returns rows
2) accepts rows calculates and generates a report
3) only write the report to a file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,95 @@
...
+                bufferedWriter.flush();
+            }
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write file: " + toFileName + ". Error: " + e);
+        } finally {
+            closeBufferWriter(toFileName);
+        }
     }
+
+    public HashMap<String, Integer> calculateStatisticResults(String fromFileName) {
```
liliia-ponomarenko, line 36: can be private

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,95 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.HashMap;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final String STRING_SEPARATOR = ",";
+    private static final String[] REPORT_COLUMNS = new String[] {"supply", "buy", "result"};
+    private static BufferedWriter bufferedWriter = null;
+    private static File file = null;
```
liliia-ponomarenko, line 15: make it a local variable

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,95 @@
...
+        if (bufferedWriter != null) {
+            try {
+                bufferedWriter.close();
+            } catch (IOException e) {
+                throw new RuntimeException("Can't close file: " + fileName + ". Error: " + e);
+            }
+        }
+    }
+
+    private void clearFile(String fileName) {
```
liliia-ponomarenko, line 77: Why do you need this method? Let’s  remove it

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,95 @@
...
+        }
     }
+
+    public HashMap<String, Integer> calculateStatisticResults(String fromFileName) {
+        HashMap<String, Integer> results = new HashMap<>();
+        file = new File(fromFileName);
+
+        try {
+            List<String> lines = Files.readAllLines(file.toPath());
+
```
liliia-ponomarenko, line 42: use try with resources construction

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
arastorhuiev: @liliia-ponomarenko I make this method for complete next test:
image
No idea what solution better, I think we have 2 variants:
1) if file exists, remove file and create new file; (I think this variant better than 2)
2) read file & write empty string;

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.HashMap;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final String STRING_SEPARATOR = ",";
```
Elena-Bruyako, line 12:
```suggestion
    private static final String COMMA = ",";
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        writeFile(fromFileName, toFileName);
+    }
+
+    private void writeFile(String fromFileName, String toFileName) {
+        HashMap<String, Integer> results = calculateStatisticResults(fromFileName);
+
+        try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
+            for (String column: REPORT_COLUMNS) {
+                bufferedWriter.write(column + "," + results.get(column) + System.lineSeparator());
```
Elena-Bruyako, line 24: Use StringBuilder to avoid creating new String during concatenation
```suggestion
                bufferedWriter.write(column + COMMA + results.get(column) + System.lineSeparator());
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,62 @@
...
+import java.nio.file.Files;
+import java.util.HashMap;
+import java.util.List;
+
 public class WorkWithFile {
+    private static final String STRING_SEPARATOR = ",";
+    private static final String[] REPORT_COLUMNS = new String[] {"supply", "buy", "result"};
+
     public void getStatistic(String fromFileName, String toFileName) {
+        writeFile(fromFileName, toFileName);
```
Elena-Bruyako, line 16: it's better to divide logic
create methods
```
read from file
calculate report
write report into file
```

Elena-Bruyako (changes requested):
Good solution, let's improve it

arastorhuiev:
@Elena-Bruyako Ty for details. I thought about splitting getReportByLines into two methods, but idk, maybe its unnecessary (or overengineering)?

Elena-Bruyako (approved):

PR #1579, opened 2024-12-07
Added method for getting statistic

SSXcorp:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
+import java.util.TreeMap;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        writeDataToFile(getDataFromFile(fromFileName), toFileName);
```
liliia-ponomarenko, line 14: You need to split the logic into three private methods and call them one by one in the public method 
1) only reads data and returns strings
2) accepts lines, calculates and forms the text of the report
3) accepts the report and only writes it to a file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,56 @@
...
+    }
+
+    private static Map<String, Integer> getDataFromFile(String fromFileName) {
```
liliia-ponomarenko, line 17: why static?

liliia-ponomarenko (changes requested):
Let’s refactor your solution!  Also, fix all comments from the common mistakes list.

src/main/java/core/basesyntax/WorkWithFile.java
SSXcorp: I made this method more flexible to read any string (not only buy or supply), but answering your question I didn't have any reasons to make it static (already fixed). I refactored code, made it simple as possible to execute only required task.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,61 @@
...
+    }
+
+    private Map<String, Integer> calculateAndFormReport(List<String> lines) {
+        int supply = 0;
+        int buy = 0;
+        int addingNumber;
+        Map<String, Integer> map = new LinkedHashMap<>();
+        for (String string : lines) {
+            addingNumber = Integer.parseInt(string.substring(string.indexOf(',') + 1));
+            if (string.contains("supply")) {
```
Elena-Bruyako, line 39: create constants for supply, buy, result and comma symbol

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/WorkWithFile.java
SSXcorp: > create constants for supply, buy, result and comma symbol

Do you mean final class fields for supply, buy Strings and comma char symbol?

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
 public class WorkWithFile {
+
```
liliia-ponomarenko, line 14:
```suggestion
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+    private final char comma = ',';
+    private final String supply = "supply";
+    private final String buy = "buy";
+    private final String result = "result";
```
liliia-ponomarenko, line 18:
```suggestion
    private static final char COMMA = ',';
    private static final String SUPPLY = "supply";
    private static final String BUY = "buy";
    private static final String RESULT = "result";
```

liliia-ponomarenko (approved):
Good job! A few minor comments ;)

PR #1603, opened 2025-01-12
done

mizuo-san:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
```
liliia-ponomarenko, line 11: You need to split the logic into three private methods and call them one by one in the public method 
1) only reads data and returns strings
2) accepts lines, calculates and forms the text of the report
3) accepts the report and only writes it to a file

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+        try {
+            writeToFile(fromFileName, toFileName);
+        } catch (IOException e) {
+            System.out.println("Unknown IOException error occurred while running " + e);
```
liliia-ponomarenko, line 15: throw RuntimeException

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+                    default:
+                        break;
+                }
+                readLines = reader.readLine();
+            }
+            result = supply - buy;
+            finalResult[0] = supply;
+            finalResult[1] = buy;
+            finalResult[2] = result;
+            return finalResult;
```
liliia-ponomarenko, line 46: The try block should contain the smallest part of the code where a potential exception can occur.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+            String readLines = reader.readLine();
+            while (readLines != null) {
+                String[] splitter = readLines.split(",");
+                switch (splitter[0]) {
+                    case "supply":
+                        supply += Integer.parseInt(splitter[1]);
+                        break;
+                    case "buy":
+                        buy += Integer.parseInt(splitter[1]);
+                        break;
```
liliia-ponomarenko, line 36: create constants for literals and magic numbers

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,66 @@
...
+            finalResult[0] = supply;
+            finalResult[1] = buy;
+            finalResult[2] = result;
+            return finalResult;
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read data from the file " + fromFileName, e);
+        }
+    }
+
+    private void writeToFile(String fromFileName, String toFileName) throws IOException {
```
liliia-ponomarenko, line 52: the same comments. Don’t  throw exceptions in the method signature

liliia-ponomarenko (changes requested):

PR #1615, opened 2025-01-25
first solution

anpiet:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+        int supply = 0;
+        int buy = 0;
+        File file = new File(fromFileName);
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(file))) {
+            String line = reader.readLine();
+            while (line != null) {
+                String[] lineWords = line.split(",");
+                String operation = lineWords[0].trim();
+                int value = Integer.parseInt(lineWords[1]);
```
slade13, line 21: The code does not handle the case where the second element of the line is not a valid integer, which could cause a `NumberFormatException`. Consider adding error handling for this scenario.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
     public void getStatistic(String fromFileName, String toFileName) {
+        int supply = 0;
+        int buy = 0;
+        File file = new File(fromFileName);
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(file))) {
+            String line = reader.readLine();
+            while (line != null) {
+                String[] lineWords = line.split(",");
+                String operation = lineWords[0].trim();
```
slade13, line 20: The code assumes that each line in the CSV file has exactly two elements separated by a comma. It would be safer to check the length of `lineWords` before accessing the second element to avoid potential `ArrayIndexOutOfBoundsException`.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+                int value = Integer.parseInt(lineWords[1]);
+                if (operation.equals("supply")) {
+                    supply += value;
+                }
+                if (operation.equals("buy")) {
```
slade13, line 25: Using `else if` instead of a second `if` statement would be more efficient since an operation cannot be both 'supply' and 'buy' at the same time.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+            throw new RuntimeException("Problem reading file", ex);
+        }
+
+        int result = supply - buy;
 
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(toFileName))) {
+            writer.write("supply," + supply + "\n");
+            writer.write("buy," + buy + "\n");
+            writer.write("result," + result);
+        } catch (Exception e) {
```
slade13, line 40: Catching a generic `Exception` is not a good practice. It is better to catch specific exceptions, such as `IOException`, to provide more meaningful error handling.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,44 @@
...
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
     public void getStatistic(String fromFileName, String toFileName) {
+        int supply = 0;
+        int buy = 0;
+        File file = new File(fromFileName);
```
slade13, line 14: The `File` object creation is unnecessary here. You can directly pass the `fromFileName` to the `FileReader` constructor: https://docs.oracle.com/javase/8/docs/api/java/io/FileReader.html#FileReader-java.lang.String-

PR #1622, opened 2025-02-04
Reads a CSV file and writes the result

Eugene-Yukhimenko:
…nce, and writes the result to a file, ensuring a newline at the end.


Eugene-Yukhimenko (commented):
Replaced the use of RemoteException with RuntimeException. Tasks were split into three separate methods and added file validation.

Eugene-Yukhimenko:
Replaced the use of RemoteException with RuntimeException. Tasks were split into three separate methods and added file validation.

PR #1648, opened 2025-03-14
First solution.

dariuszb:


dariuszb:
I have a problem with split code between smallest part

PR #1694, opened 2025-07-04
Implement getStatistic method for reading and writing CSV report

MaritaPasha:


MaritaPasha:
There is no header line in my test file. When I add a blank line at the beginning (to simulate a header), the tests fail. Therefore, I did not include a header, and my code processes the file starting from the first actual data line.

PR #1731, opened 2025-09-02
Implement getStatistic method to read CSV file, calculate statistics and write report

tsuki99:
…enerate report

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -10,6 +10,7 @@ public class WorkWithFile {
     private static final String COMMA = ",";
     private static final String BUY = "buy";
     private static final String SUPPLY = "supply";
+    private static final String RESULT = "result";
     private static final int INITIAL_SUM = 0;
 
     public void getStatistic(String fromFileName, String toFileName) {
@@ -35,8 +36,12 @@ public class WorkWithFile {
             throw new RuntimeException("Can't read the file", e);
         }
 
+        writeToFile(toFileName, getReport(sumBuy, sumSupply));
+    }
+
+    private void writeToFile(String toFileName, String report) {
         try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(toFileName))) {
-            bufferedWriter.write(getReport(sumBuy, sumSupply));
+            bufferedWriter.write(report);
         } catch (IOException e) {
             throw new RuntimeException("Can't write to the file");
         }
@@ -44,9 +49,19 @@ public class WorkWithFile {
 
     private String getReport(int sumBuy, int sumSupply) {
         int result = sumSupply - sumBuy;
+        StringBuilder report = new StringBuilder();
+        report.append(SUPPLY)
+                .append(COMMA)
+                .append(sumSupply)
+                .append(System.lineSeparator())
+                .append(BUY)
+                .append(COMMA)
+                .append(sumBuy)
+                .append(System.lineSeparator())
+                .append(RESULT)
+                .append(COMMA)
+                .append(result);
 
-        return SUPPLY + COMMA + sumSupply + System.lineSeparator()
-                + BUY + COMMA + sumBuy + System.lineSeparator()
-                + "result" + COMMA + result;
+        return report.toString();
     }
 }
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -8,36 +8,47 @@ import java.io.IOException;
 
 public class WorkWithFile {
     private static final String COMMA = ",";
+    private static final String SPACE = " ";
     private static final String BUY = "buy";
     private static final String SUPPLY = "supply";
     private static final String RESULT = "result";
     private static final int INITIAL_SUM = 0;
 
     public void getStatistic(String fromFileName, String toFileName) {
-        int sumBuy = readFromFile(fromFileName, BUY);
-        int sumSupply = readFromFile(fromFileName, SUPPLY);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,52 @@
...
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.FileReader;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String COMMA = ",";
+    private static final String BUY = "buy";
+    private static final String SUPPLY = "supply";
```
Elena-Bruyako, line 13: make `result` as a constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,52 @@
...
+            bufferedWriter.write(getReport(sumBuy, sumSupply));
+        } catch (IOException e) {
+            throw new RuntimeException("Can't write to the file");
+        }
+    }
+
+    private String getReport(int sumBuy, int sumSupply) {
+        int result = sumSupply - sumBuy;
 
+        return SUPPLY + COMMA + sumSupply + System.lineSeparator()
```
Elena-Bruyako: use StringBuilder to create report

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,52 @@
...
+import java.io.IOException;
+
 public class WorkWithFile {
+    private static final String COMMA = ",";
+    private static final String BUY = "buy";
+    private static final String SUPPLY = "supply";
+    private static final int INITIAL_SUM = 0;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int sumBuy = INITIAL_SUM;
```
Elena-Bruyako: divide method into 3 methods:
```
read from file 
calculate report
write into file
```

Elena-Bruyako (changes requested):
Good job! Let's improve your solution

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,67 @@
...
+    private static final String BUY = "buy";
+    private static final String SUPPLY = "supply";
+    private static final String RESULT = "result";
+    private static final int INITIAL_SUM = 0;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int sumBuy = INITIAL_SUM;
+        int sumSupply = INITIAL_SUM;
+
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader(fromFileName))) {
```
Elena-Bruyako, line 27: this logic should be in `read from file` method

Elena-Bruyako (changes requested):
one comment

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
+    private static final int INITIAL_SUM = 0;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int sumBuy = readFromFile(fromFileName, BUY);
+        int sumSupply = readFromFile(fromFileName, SUPPLY);
+
+        writeToFile(toFileName, getReport(sumBuy, sumSupply));
+    }
+
+    private int readFromFile(String fromFileName, String keyWord) {
```
Elena-Bruyako:
```suggestion
    private int[] readFromFile(String fromFileName) {
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,71 @@
...
 public class WorkWithFile {
+    private static final String COMMA = ",";
+    private static final String BUY = "buy";
+    private static final String SUPPLY = "supply";
+    private static final String RESULT = "result";
+    private static final int INITIAL_SUM = 0;
+
     public void getStatistic(String fromFileName, String toFileName) {
+        int sumBuy = readFromFile(fromFileName, BUY);
+        int sumSupply = readFromFile(fromFileName, SUPPLY);
```
Elena-Bruyako: it should be 3 separate methods
each of responsible only for one particular task (logic) 
it's Single responsibility principle 
```suggestion
        int[] output = readFromFile(fromFileName);
        String result = getReport(output);
        writeToFile(result, toFileName);
```

Elena-Bruyako (changes requested):
see comments

src/main/java/core/basesyntax/WorkWithFile.java
tsuki99: @Elena-Bruyako 
I think it's a bad solution to solve the problem this way. You suggest dividing the logic, but this will only add a few unnecessary actions like creating additional arrays and going through them, instead of immediately reading the information from the file and passing it to separate variables. I think the best solution would be to rename the method to getSumByKeyword, this way we don't do unnecessary operations, but immediately work directly with the file. Taking content and processing it several times is bad practice in my opinion.

Or we can create a getData method, not "readFromFile", because if we read, the method should not return anything, and then what's the point of it. Then this method (getData) collects them and passes them to a separately created "getSum", which we pass to the corresponding variables. But even so, this will be the wrong solution because we create unnecessary operations and arrays.

Elena-Bruyako: Уяви ситуацію, коли в тебе буде не дві операції. Для кожної з них ти будеш викликати один і той самий метод. Краще зробити певну логіку в методі, за яку він відповідає та передати результат

Elena-Bruyako (changes requested):
see comment

PR #1766, opened 2025-10-27
Final commit

PiotrHic:


MateuszRostek (changes requested):
Before creating a PR please make sure that you are not making the common mistakes attached at the end of the task description:
[Try to avoid these common mistakes, while solving task](https://github.com/mate-academy/jv-work-with-file/blob/master/checklist.md)
(Divide program logic)

PR #1830, opened 2026-03-02
WorkWithFile first solution

mate-student26:


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
 package core.basesyntax;
 
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
 public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+    public String[] getStatistic(String fromFileName, String toFileName) {
```
slade13, line 9: Wrong return type - the expected result is writing to file, not returning a `String[]`.
```suggestion
    public void getStatistic(String fromFileName, String toFileName) {
```

slade13, line 9: Your method does everything:

1. validates input
2. reads file
3. parses data
4. calculates statistics
5. builds report
6. writes to file
7. handles exceptions

This method is ~50 lines long and does too many things.

The task explicitly says:

> Methods should be short (5–15 lines)
> One method -> one responsibility

Example clean structure:
```
    validateFiles(fromFileName, toFileName);
    List<String> lines = readFromFile(fromFileName);
    Statistic statistic = calculateStatistic(lines);
    String report = createReport(statistic);
}
```

Then:

`readFromFile()` - only reading
`calculateStatistic() `- only calculations
`createReport()` - only formatting
`writeToFile()` - only writing

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Can't read input file: " + fromFileName, e);
+        } catch (NumberFormatException | ArrayIndexOutOfBoundsException e) {
+            throw new RuntimeException("Bad input format in file: " + fromFileName, e);
+        }
+
+        try (BufferedWriter writer = Files.newBufferedWriter(Path.of(toFileName))) {
+            writer.write(result.toString());
+        } catch (Exception e) {
+            throw new NullPointerException();
```
slade13, line 53: This block is very problematic:

- You ignore original exception
- You lose stacktrace
- You throw wrong exception type
- You provide no message
- This violates: "Don't ignore exceptions"

Correct version:

```suggestion
    catch (IOException e) {
        throw new RuntimeException("Can't write data to file: " + toFileName, e);
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
 
+                if (operationType.equals("buy")) {
+                    buyAmount += amount;
+                } else if (operationType.equals("supply")) {
+                    supplyAmount += amount;
+                }
+            }
+
+            int difference = supplyAmount - buyAmount;
+            result.append("supply,")
```
slade13, line 35: Use constants instead of magic strings:
```
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
 public class WorkWithFile {
-    public void getStatistic(String fromFileName, String toFileName) {
+    public String[] getStatistic(String fromFileName, String toFileName) {
+        if (fromFileName == null || toFileName == null) {
+            throw new NullPointerException("No file to read and/or no file to write");
```
slade13, line 11: Throwing `NullPointerException` manually is not good practice, it's better to use for example `IllegalArgumentException`.

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+
+        StringBuilder result = new StringBuilder();
+        final String ls = System.lineSeparator();
+
+        try {
+            String[] lines = Files.readAllLines(Path.of(fromFileName)).toArray(new String[0]);
+            int buyAmount = 0;
+            int supplyAmount = 0;
+
+            for (int i = 0; i < lines.length; i++) {
```
slade13, line 22: Instead using index loop use enhanced for:

```suggestion
            for (String line : lines)
```

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,58 @@
...
+                String[] splittedLine = lines[i].split(",");
```
slade13, line 23: You assume:

- Exactly 2 elements
- Proper format

It's better to validate input:
```
if (splittedLine.length != 2) {
    throw new RuntimeException("Invalid line format: " + line);
}
```

slade13 (changes requested):
Please review my comments and update your code accordingly.

src/main/java/core/basesyntax/WorkWithFile.java
mate-student26: Hi, 
According to this table 
image

RuntimeException is not under IOException but directly under Exception.
How can we explain tthis on example above ?
Shoudl be here "checked exception" or "unchecked exception" ?
Trying now, by reviewing EXCEPTION module study to understand more when/where/which exception shoudl be used ?

PR #1854, opened 2026-05-04
Add solution for work-with-file task

DanyilShkodiuk:
- Read CSV data from file using BufferedReader
- Calculate supply, buy totals and result
- Write report to output file using BufferedWriter

### DIFF
--- a/src/main/java/core/basesyntax/WorkWithFile.java
+++ b/src/main/java/core/basesyntax/WorkWithFile.java
@@ -12,6 +12,7 @@ public class WorkWithFile {
     private static final String SUPPLY = "supply";
     private static final String BUY = "buy";
     private static final String SEPARATOR = ",";
+    private static final String RESULT = "result";
 
     public void getStatistic(String fromFileName, String toFileName) {
         List<String> lines = readFile(fromFileName);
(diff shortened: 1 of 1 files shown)


src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,60 @@
...
+            if (parts[0].equals(SUPPLY)) {
+                supply += amount;
+            } else {
+                buy += amount;
+            }
+        }
+        int result = supply - buy;
+        return SUPPLY + SEPARATOR + supply + "\n"
+                + BUY + SEPARATOR + buy + "\n"
+                + "result" + SEPARATOR + result;
```
Elena-Bruyako: make "result" as a constant

src/main/java/core/basesyntax/WorkWithFile.java
```diff
@@ -1,7 +1,60 @@
...
+            String[] parts = line.split(SEPARATOR);
+            int amount = Integer.parseInt(parts[1]);
+            if (parts[0].equals(SUPPLY)) {
+                supply += amount;
+            } else {
+                buy += amount;
+            }
+        }
+        int result = supply - buy;
+        return SUPPLY + SEPARATOR + supply + "\n"
```
Elena-Bruyako: use StringBuilder

Elena-Bruyako: use System.lineSeparator() instead of \n

Elena-Bruyako (changes requested):
see comments

Elena-Bruyako (approved):

